PR #14, opened 2026-03-14
Steel thread: scaffold issues service + CLI with ticket CRUD and E2E tests

### LINKED ISSUE #5: Steel thread: scaffold service + CLI with ticket CRUD and E2E tests
## Description

Set up both `services/issues/` and `services/issues-cli/` with the full vertical slice working end-to-end. This includes the complete Prisma schema (all models — no schema changes needed in later phases), Apollo Server, CLI with graphql-request, and E2E test infrastructure.

### What gets built

**Service (`services/issues/`):**
- package.json, tsconfig, .env.example, .gitignore
- Prisma schema with all 6 models (User, Ticket, Label, TicketLabel, Comment, BlockRelation) and enums (TicketState, Priority)
- Seed file for 6 default labels (bug, enhancement, accessibility, security, ux, performance)
- Apollo Server with ticket CRUD resolvers:
  - `createTicket` (with optional labels, storyPoints, priority)
  - `ticket(id)` query with all relations
  - `tickets` query with cursor pagination and filters (state, label, assignee, blocked, priority)
  - `updateTicket` (title, description, acceptanceCriteria, storyPoints, priority)
  - No `deleteTicket` — tickets are closed, not deleted

**CLI (`services/issues-cli/`):**
- package.json, tsconfig
- GraphQL client setup (graphql-request)
- Config management (API URL, auth token)
- Ticket commands: create, view, list, update

**Tests & Docs:**
- E2E test infrastructure: helpers.ts (test server, CLI runner, test auth)
- ticket-management.e2e.ts, priority-estimation.e2e.ts
- User story docs in `docs/user-stories/` (all 8 files)

## Acceptance Criteria

- [ ] `services/issues/` scaffolded with package.json, tsconfig.json, .env.example, .gitignore
- [ ] Prisma schema with all 6 models, TicketState + Priority enums
- [ ] `prisma migrate dev` creates initial migration successfully
- [ ] `prisma/seed.ts` creates 6 default labels (idempotent via upsert)
- [ ] `npm run dev` starts Apollo Server on :4000
- [ ] `services/issues-cli/` scaffolded with package.json, tsconfig.json
- [ ] CLI can create, view, list, and update tickets against the running server
- [ ] E2E tests pass: ticket CRUD and priority/estimation scenarios
- [ ] Full stack exercised: CLI → graphql-request → Apollo Server → Prisma → SQLite
- [ ] User story docs written in `docs/user-stories/` (all 8 files)

dokipen:
## Summary

- Scaffold `services/issues/` with Apollo Server, Prisma (SQLite), and ticket CRUD resolvers (create, view, list with cursor pagination + filters, update)
- Scaffold `services/issues-cli/` with commander, graphql-request, and ticket commands (create, view, list, update)
- Full Prisma schema with all 6 models (User, Ticket, Label, TicketLabel, Comment, BlockRelation) and enums — no schema changes needed in later phases
- E2E test infrastructure with 18 passing tests exercising the full stack: CLI → graphql-request → Apollo Server → Prisma → SQLite
- 8 user story documents in `docs/user-stories/`

Fixes #5

## Test plan

- [x] `services/issues/` typechecks (`tsc --noEmit`)
- [x] `services/issues-cli/` typechecks (`tsc --noEmit`)
- [x] Prisma migration + seed run successfully
- [x] Apollo Server starts on :4000
- [x] 18 E2E tests pass (ticket-management + priority-estimation)
- [x] Full stack exercised end-to-end
- [ ] Code review


dokipen (commented):
## Security Review

Reviewed: `services/issues/` (Apollo Server + Prisma) and `services/issues-cli/` (CLI client). Auth is intentionally absent (Phase 5) and is not flagged here.

---

### Dependency Audit

Both `npm audit` results are clean — zero vulnerabilities across all 311 + 128 resolved packages. Dependencies are current.

One note: `jsonwebtoken` is listed as a runtime dependency in `services/issues/package.json` but is not yet used anywhere in the codebase. That is fine as a forward-looking placeholder, but it should not be imported before auth is actually implemented (unused crypto dependencies are a small but real supply-chain surface).

---

### Risk Assessment

| # | Finding | Severity | Location | Recommendation |
|---|---------|----------|----------|----------------|
| 1 | Unbounded pagination `first` parameter | Warning | `services/issues/src/schema/resolvers/ticket.ts:26` | Clamp `first` to a maximum (e.g., 100). Currently a caller can request an arbitrarily large result set, which is a DoS vector once the service is network-exposed. |
| 2 | `state` and `priority` enum values not validated server-side | Warning | `services/issues/src/schema/resolvers/ticket.ts:31–36` and `prisma/schema.prisma:34–35` | The GraphQL schema correctly declares `TicketState` and `Priority` enums, so the Apollo layer rejects invalid values on `createTicket`/`updateTicket` inputs. However, `state` and `priority` are stored as raw strings in SQLite with no DB-level constraint. If data is ever written by anything other than the GraphQL layer (direct Prisma scripts, migrations, tests), invalid values can enter the DB silently. Add an application-layer allowlist check in the resolvers (a simple `VALID_STATES.includes(state)` guard) as defence-in-depth. |
| 3 | `auth.json` written with default filesystem permissions | Warning | `services/issues-cli/src/config.ts:54–55` | `writeFileSync(AUTH_FILE, ...)` creates the file with the process umask (typically `0o644`), meaning other users on the same machine can read the bearer token. Change the write call to use mode `0o600`: `writeFileSync(AUTH_FILE, content, { encoding: "utf-8", mode: 0o600 })`. |
| 4 | `JWT_SECRET` placeholder in `.env.example` | Suggestion | `services/issues/.env.example:2` | The value `"change-me-in-production"` is fine for an example file, but add a comment explicitly warning that the secret must be replaced and should be at least 256 bits of random data. This prevents copy-paste deployments with the placeholder value. |
| 5 | `graphql-tag` listed as a runtime dependency but unused | Suggestion | `services/issues/package.json:24` | The server uses string literal type-defs, not `gql` tagged templates. `graphql-tag` appears to be an unused dependency. Remove it to reduce the attack surface. |
| 6 | `services/issues-cli/.gitignore` does not exclude `auth.json` | Suggestion | `services/issues-cli/.gitignore` | The config dir (`~/.issues-cli/`) is outside the repo, so accidental commit of `auth.json` is unlikely. Still, adding `auth.json` to the ignore file (or a note in the README) makes the intent explicit for contributors who might move config in-tree later. |

---

### Detailed Notes

**Finding 1 — Unbounded pagination (DoS)**

```ts
// resolvers/ticket.ts:26
const { first = 20, after } = args;
// ...
take: first + 1,
```

There is no upper bound on `first`. A request with `first: 1000000` will issue a single SQLite query for a million rows. Before auth lands, the service is presumably only reachable locally, so the practical risk is low now — but this should be fixed before the service accepts network traffic.

```ts
const MAX_PAGE_SIZE = 100;
const take = Math.min(first, MAX_PAGE_SIZE) + 1;
```

**Finding 2 — Enum stored as unconstrained string**

The Prisma schema comment itself acknowledges this: "SQLite does not support enums natively — we store them as strings and validate at the application layer." The GraphQL input types do enforce the enum, so the mutation path is safe. The risk is any out-of-band write path. A resolver-level guard (or a Prisma middleware/`$extends` validator) would close this gap.

**Finding 3 — Token file permissions**

This is the most actionable pre-auth finding. The fix is a one-line change to `setAuthToken`:

```ts
writeFileSync(AUTH_FILE, JSON.stringify(auth, null, 2) + "\n", { encoding: "utf-8", mode: 0o600 });
```

---

### Summary

- **Overall risk level**: Low (pre-auth, local-only deployment)
- **No critical findings** — no committed secrets, no SQL injection (Prisma parameterises all queries), no hardcoded credentials, no command injection vectors, clean dependency audit
- **Prioritised remediation before auth phase**:
  1. Fix `auth.json` file permissions (Finding 3) — trivial one-liner
  2. Add pagination limit (Finding 1) — prevents DoS when service becomes network-accessible
  3. Add server-side enum validation (Finding 2) — defence-in-depth for the string-typed SQLite columns

dokipen (commented):
## Code Review

Overall this is solid scaffold work. The architecture is clean, the E2E test infrastructure is genuinely good, and the GraphQL / Prisma design is sensible. I have a few issues worth addressing before merge, and some suggestions for later phases.

---

### Critical (must fix)

**1. N+1 query problem in field resolvers**
`services/issues/src/schema/resolvers/ticket.ts`, lines 163-198

Every field on `Ticket` (`labels`, `comments`, `assignee`, `blocks`, `blockedBy`) issues its own SQL query. When `tickets` returns N results, listing tickets fires 5N+1 database queries. This will be noticeably slow even at 20 rows.

The immediate fix for the `assignee` field is to add `include: { assignee: true }` to the `findMany` query (the relation is on the same row, so no join is even needed — Prisma just reads `assigneeId` back). For `labels`, `comments`, `blocks`, and `blockedBy`, a DataLoader is the conventional fix, but at minimum the `assignee` case should be fixed now since it is a simple scalar on the `Ticket` row.

Concretely, `Ticket.assignee` on line 179 re-queries the DB for a user whose `id` is already available on `parent`. The entire resolver can be replaced by including `assignee` in the parent query.

**2. `updateTicket` resolver does not verify the ticket exists**
`services/issues/src/schema/resolvers/ticket.ts`, lines 133-158

`prisma.ticket.update` throws a Prisma `P2025` "record not found" error when the `id` does not exist. That error surfaces as a raw Prisma exception rather than a clean `"Ticket not found"` GraphQL error. The resolver should either call `findUnique` first, or wrap the update in a try/catch that maps `P2025` to a user-facing message. The `createTicket` resolver has the same gap for invalid `assigneeId` or `labelIds` references.

**3. No input validation on enums at the resolver layer**
`services/issues/src/schema/resolvers/ticket.ts`, lines 31-37, 116, 153

Because SQLite does not support native enums and the Prisma schema stores `state` and `priority` as plain strings, any string value can be written to the database. The GraphQL schema enums provide client-side validation, but a direct API call with `priority: "BOGUS"` will write `"BOGUS"` to SQLite with no error. The resolver should validate enum inputs against the allowed values before writing.

---

### Warnings (should fix)

**4. Type-safety bypassed with `as` casts**
Files:
- `services/issues/src/schema/resolvers/ticket.ts` line 75: `queryArgs as Parameters<typeof prisma.ticket.findMany>[0]`
- `services/issues/src/schema/resolvers/ticket.ts` line 129: `data as Parameters<typeof prisma.ticket.create>[0]["data"]`
- `services/issues/src/schema/resolvers/ticket.ts` line 156: `data` passed directly to `prisma.ticket.update`

These casts hide real type mismatches. Building query objects as `Record<string, unknown>` then casting to Prisma's typed arguments defeats strict mode. The Prisma-generated types are specific enough to be used directly; the resolver should use typed intermediates (e.g., `Prisma.TicketWhereInput`, `Prisma.TicketCreateInput`) rather than loose record types.

**5. `TicketLabel` model is missing a primary key**
`services/issues/prisma/schema.prisma`, lines 57-65

The `TicketLabel` join table has a `@@unique([ticketId, labelId])` constraint but no `@id` field. Prisma requires either `@id` or `@@id` on every model. This may work with SQLite today but will fail migrations on PostgreSQL and will break Prisma's `findUnique` / `update` methods on the join table. Add `@@id([ticketId, labelId])` and remove the `@@unique` (they are equivalent, but `@@id` is the correct declaration for a composite primary key).

**6. Race condition in `getFreePort`**
`services/issues-cli/test/e2e/helpers.ts`, lines 15-29

The helper finds a free port by listening, records it, then closes the socket before the server binds it. Another process can claim the port in that window. For test isolation this is acceptable but is a known pattern that occasionally causes flaky test failures in CI with parallel test runners. Using `PORT=0` and reading the actual bound port from the server's `listen` callback is the reliable alternative.

**7. `graphql-tag` dependency is unused in `services/issues`**
`services/issues/package.json`, line 24

`graphql-tag` is listed as a production dependency but the server only uses string type definitions — there is no `gql` import anywhere in the service. Remove it to keep the dependency footprint clean.

**8. `jsonwebtoken` installed but authentication is not implemented**
`services/issues/package.json`, line 22

`jsonwebtoken` (and `@types/jsonwebtoken`) are declared as dependencies but nothing in the current source uses them. User story 07 (auth) is deferred, so this is fine, but the unused dependency should either be removed until auth is implemented or at minimum moved to `devDependencies` with a comment explaining it is reserved.

**9. `isBlocked: false` filter produces a surprising result**
`services/issues/src/schema/resolvers/ticket.ts`, lines 53-60

When `isBlocked` is passed as `false`, the resolver adds `blockedBy: { none: {} }` to the query, returning only tickets with _no_ blockers. This behaviour is not mentioned in the CLI help text for `--blocked` (which is a flag, not an option with a boolean value). The CLI can only pass `isBlocked: true`, so `false` is never sent — but the schema exposes it and the behaviour should be documented or the schema should limit it to a `blocked: Boolean` flag-style argument.

---

### Suggestions (nice to have)

**10. Test ordering dependency is fragile**
Files: `services/issues-cli/test/e2e/ticket-management.e2e.ts`, `priority-estimation.e2e.ts`

Tests share `createdTicketId` / `fullTicketId` variables that are populated by earlier tests in the same `describe` block. If the creating test fails, subsequent tests that use the ID will fail with a confusing error. The `vitest.config.ts` comment acknowledges this ("may depend on order"), but the fragility could be reduced by giving each test full setup or by using `beforeAll` to create fixture data.

**11. Spinner output mixed into stdout assertions**
Multiple test files, e.g. `ticket-management.e2e.ts` line 25:

```ts
const output = result.stdout + result.stderr;
```

This works because `NO_COLOR=1` disables chalk, but `ora` still writes to stderr. The comment on line 24 explains it, which is good. Consider routing spinner output through stderr consistently in the CLI (it already does this via `ora`) and asserting data-bearing output through `result.stdout` alone, which is what the spinner-success message and ticket data tests already do correctly.

**12. `Ticket` field resolver for `createdAt`/`updatedAt` returns a raw Prisma `Date`, not a `DateTime` scalar**

The `DateTime` scalar from `graphql-scalars` is registered at the server level, but the field resolvers return raw `Ticket` objects whose `createdAt`/`updatedAt` are JavaScript `Date` instances. `DateTimeResolver` from `graphql-scalars` serializes `Date` objects correctly, so this works at runtime — but it is worth noting the implicit dependency: if the scalar registration were ever removed, the dates would silently serialize as strings via `.toISOString()`.

**13. `--labels` option on `ticket create` accepts label IDs, not names**
`services/issues-cli/src/commands/ticket.ts`, line 235:

```
.option("--labels <ids>", "Comma-separated label IDs")
```

Requiring callers to know internal CUIDs is a poor UX. Future work should either expose a `labels list` command to discover IDs, or accept label names and resolve them server-side (a `labelNames` input field on `CreateTicketInput`).

---

### Summary

| # | Severity | Location |
|---|----------|----------|
| 1 | Critical | `resolvers/ticket.ts` — N+1 queries on list |
| 2 | Critical | `resolvers/ticket.ts` — no not-found handling in `updateTicket` |
| 3 | Critical | `resolvers/ticket.ts` — enum values not validated before DB write |
| 4 | Warning | `resolvers/ticket.ts` — type-safety bypassed with `as` casts |
| 5 | Warning | `schema.prisma` — `TicketLabel` missing `@@id` |
| 6 | Warning | `helpers.ts` — TOCTOU race on free-port detection |
| 7 | Warning | `services/issues/package.json` — unused `graphql-tag` dep |
| 8 | Warning | `services/issues/package.json` — unused `jsonwebtoken` dep |
| 9 | Warning | `resolvers/ticket.ts` — undocumented `isBlocked: false` behaviour |

Items 1–3 and 5 should be addressed before merge; the rest can follow in subsequent phases.

REVIEW STATUS: CHANGES NEEDED
- Resolver: add not-found handling to updateTicket and validate enum inputs before writing to SQLite
- Resolver: fix assignee N+1 (include in parent query); plan DataLoader for labels/comments/blocks
- Prisma schema: replace `@@unique([ticketId, labelId])` with `@@id([ticketId, labelId])` on TicketLabel
- Dependencies: remove unused `graphql-tag` and `jsonwebtoken` (or defer to devDependencies)

dokipen (commented):
## Review Fixes Applied

Addressed findings from code review and security review:

**Critical fixes:**
- `updateTicket` now returns a clean `Ticket not found` error instead of raw Prisma P2025
- Server-side enum validation for priority values (defense-in-depth against direct API access)

**Security fixes:**
- `auth.json` now written with `0o600` permissions (owner-only)
- Pagination capped at 100 items max

**Cleanup:**
- Removed unused deps: `graphql-tag`, `jsonwebtoken`, `@types/jsonwebtoken`
- `TicketLabel` uses `@@id` instead of `@@unique` (PostgreSQL compat)
- `prisma/migrations/` no longer gitignored (needed for CI/Docker)

**Deferred (tracked for future phases):**
- N+1 queries / DataLoader — acceptable for steel thread, optimize in later phase
- TOCTOU port race in test helpers — acceptable for test infrastructure
- Type-safety casts — refactoring scope

All 18 E2E tests still pass after fixes.

PR #33, opened 2026-03-15
Add FSM state transitions, blocking relationships, and guards

### LINKED ISSUE #7: FSM state transitions and blocking with guards + CLI + E2E
## Description

Implement the ticket state machine (BACKLOG → REFINED → IN_PROGRESS → CLOSED with valid transitions and demotions). Add blocking relationship mutations. Guard: blocked tickets cannot transition to IN_PROGRESS.

### State Machine
```
BACKLOG     → [REFINED]
REFINED     → [IN_PROGRESS, BACKLOG]
IN_PROGRESS → [CLOSED, REFINED]
CLOSED      → [BACKLOG]
```

Data-driven transitions map (no xstate — 4 states doesn't warrant it).

### Guard
Tickets with open blockers cannot transition to IN_PROGRESS. Error includes blocker count.

### GraphQL Operations
- `transitionTicket(id, to)` mutation
- `addBlockRelation(blockerId, blockedId)` mutation
- `removeBlockRelation(blockerId, blockedId)` mutation
- `blocks` and `blockedBy` field resolvers on Ticket type

### CLI Commands
- `issues ticket transition <id> --to <STATE>`
- `issues block add --blocker <id> --blocked <id>`
- `issues block remove --blocker <id> --blocked <id>`

## Acceptance Criteria

- [ ] `transitionTicket(id, to)` validates all transitions (6 valid, rejects invalid)
- [ ] `addBlockRelation`, `removeBlockRelation` mutations work
- [ ] `blocks` and `blockedBy` field resolvers on Ticket type
- [ ] Blocked tickets get error when transitioning to IN_PROGRESS
- [ ] CLI commands: `issues ticket transition`, `issues block add/remove`
- [ ] Unit tests: fsm.test.ts (all valid/invalid transitions, blocker guard)
- [ ] E2E tests pass: workflow-states.e2e.ts
- [ ] E2E tests pass: blocking.e2e.ts

dokipen:
## Summary
- Implement ticket state machine (BACKLOG → REFINED → IN_PROGRESS → CLOSED) with data-driven transitions map in `src/fsm/`
- Add blocker guard: blocked tickets cannot transition to IN_PROGRESS until all blockers are CLOSED
- Add `transitionTicket`, `addBlockRelation`, `removeBlockRelation` GraphQL mutations
- Add CLI commands: `ticket transition <id> --to <STATE>`, `block add`, `block remove`
- Add 16 unit tests for FSM logic and 25 E2E tests (workflow-states + blocking)

Fixes #7

## Test plan
- [x] 16 unit tests pass (`fsm.test.ts` — all valid/invalid transitions, same-state, blocker guard)
- [x] 13 E2E tests pass (`workflow-states.e2e.ts` — full transition lifecycle, invalid transitions, guard enforcement)
- [x] 12 E2E tests pass (`blocking.e2e.ts` — add/remove relations, self-block rejection, idempotency, guard integration)
- [x] All 57 E2E tests pass (no regressions in existing test suites)
- [x] TypeScript compiles cleanly

### DIFF
--- a/services/issues/src/fsm/fsm.test.ts
+++ b/services/issues/src/fsm/fsm.test.ts
@@ -65,10 +65,15 @@ describe("checkBlockerGuard", () => {
     expect(result).toEqual({ allowed: true });
   });
 
-  it("allows when all blockers are closed", async () => {
+  it("queries only for unresolved (non-CLOSED) blockers", async () => {
     const prisma = mockPrisma(0);
-    const result = await checkBlockerGuard("ticket-1", prisma);
-    expect(result).toEqual({ allowed: true });
+    await checkBlockerGuard("ticket-1", prisma);
+    expect(prisma.blockRelation.count).toHaveBeenCalledWith({
+      where: {
+        blockedId: "ticket-1",
+        blocker: { state: { not: "CLOSED" } },
+      },
+    });
   });
 
   it("rejects when some blockers are not closed", async () => {
--- a/services/issues/src/schema/resolvers/ticket.ts
+++ b/services/issues/src/schema/resolvers/ticket.ts
@@ -332,7 +332,7 @@ export const ticketResolvers = {
         where: { blockerId_blockedId: { blockerId, blockedId } },
       });
 
-      return prisma.ticket.findUnique({ where: { id: blockedId } });
+      return prisma.ticket.findUniqueOrThrow({ where: { id: blockedId } });
     },
   },
 


dokipen (commented):
## Code Review: PR #33 — FSM State Transitions, Blocking, and Guards

All 16 unit tests pass and both TypeScript type-checks are clean. The overall implementation is solid and consistent with the existing codebase patterns. I have a few findings to share.

---

### Critical

None.

---

### Warnings

**1. `removeBlockRelation` returns a nullable `Ticket` but the GraphQL schema declares `Ticket!` (non-null)**

`services/issues/src/schema/resolvers/ticket.ts`, line 333

```ts
return prisma.ticket.findUnique({ where: { id: blockedId } });
```

`prisma.ticket.findUnique` returns `Ticket | null`. The mutation is declared as `removeBlockRelation(...): Ticket!` in the schema, so returning `null` here would cause a GraphQL runtime error. The ticket existence is confirmed a few lines earlier (the `blockRelation` record is fetched by a composite key that includes `blockedId`), but from a type-safety standpoint the `findUnique` could still return `null` — e.g. if there is a race condition or a data-integrity bug. This should either use `findUniqueOrThrow` or add an explicit null-check and throw.

**2. Duplicate test case in `fsm.test.ts` — "allows when all blockers are closed" is identical to "allows when no blockers exist"**

`services/issues/src/fsm/fsm.test.ts`, lines 62-72

Both test cases mock `prisma.blockRelation.count` returning `0` and assert `{ allowed: true }`. The second test is supposed to cover the case where blockers exist but are all `CLOSED`, but the mock is set up exactly the same way as the first. The mock would need to verify that the Prisma query includes the `state: { not: "CLOSED" }` filter to meaningfully distinguish this case. In its current form the second test adds no coverage.

**3. No cycle detection for blocking relationships**

`services/issues/src/schema/resolvers/ticket.ts`, lines 295-317

A → B → A circular chains are not prevented. A ticket cannot block itself (checked), but there is nothing stopping A from blocking B while B already blocks A. The `IN_PROGRESS` guard would fire on both, creating a permanent deadlock with no recovery path. Consider adding a transitive cycle check before creating a `blockRelation`, or at minimum documenting this as a known limitation.

---

### Suggestions

**4. `TicketState` type in `transitions.ts` is defined but never used as a type constraint**

`services/issues/src/fsm/transitions.ts`, lines 1-8

`TicketState` is exported but `ALLOWED_TRANSITIONS` is typed as `Record<string, string[]>` rather than `Record<TicketState, TicketState[]>`. Using the narrower type would give compile-time errors if a state is added or removed without updating the transition map.

**5. `transition` subcommand help text lists `CLOSED` as a valid `--to` value but does not mention `REFINED`**

`services/issues-cli/src/commands/ticket.ts`, added lines around line 519

```ts
.requiredOption("--to <state>", "Target state (BACKLOG, REFINED, IN_PROGRESS, CLOSED)")
```

The description is correct — just noting this is fine to leave as-is, but a future improvement would be to derive the list dynamically from `ALLOWED_TRANSITIONS` to prevent it drifting.

**6. `addBlockRelation` performs two separate `findUnique` calls that could be replaced with a single query**

`services/issues/src/schema/resolvers/ticket.ts`, lines 304-308

Both the blocker and blocked tickets are fetched individually before the upsert. Since both IDs come from user input, they could be verified in a single `prisma.ticket.count({ where: { id: { in: [blockerId, blockedId] } } })` check. Minor, but reduces DB round trips.

**7. E2E test ID extraction uses `!` non-null assertion without a fallback error message**

Files: `workflow-states.e2e.ts` and `blocking.e2e.ts` (multiple locations)

```ts
ticketId = idMatch![1];
```

If the regex does not match, `idMatch!` will throw a cryptic `Cannot read properties of null` rather than pointing to the test that failed. The `expect(idMatch).toBeTruthy()` assertion on the line before will catch this — but the assertion failure message will be unhelpful. Prefer `expect(idMatch, "ticket ID not found in output").toBeTruthy()` to produce a clearer failure.

---

### Summary

The implementation is clean and well-structured. The FSM logic is correctly separated, the blocker guard is straightforward, and the test coverage is good. The two issues worth addressing before merge are the nullable return type on `removeBlockRelation` (#1) and the non-coverage of the second `checkBlockerGuard` test case (#2). The cycle-detection gap (#3) is worth either fixing or explicitly documenting as out of scope.

REVIEW STATUS: CHANGES NEEDED
- `removeBlockRelation` resolver should use `findUniqueOrThrow` or an explicit null-check to satisfy `Ticket!` return type
- Duplicate `checkBlockerGuard` unit test should be fixed to actually verify the "all blockers are CLOSED" path
- Circular blocking chain prevention should be added or documented as a known limitation

PR #35, opened 2026-03-15
Sanitize ExtraArgs to prevent shell injection

### LINKED ISSUE #32: Agent service: sanitize ExtraArgs to prevent shell injection
## Description
Discovered during PR #29 security review. `ExtraArgs` from the gRPC caller is joined into a shell string and sent to tmux via `send-keys`, allowing arbitrary command execution.

## Acceptance Criteria
- [ ] Sanitize or shell-escape ExtraArgs before sending to tmux
- [ ] Consider using `tmux send-keys` with individual args instead of shell string
- [ ] Add test for injection attempt

## Notes
Phase 1 is local-only (no auth), so the blast radius is limited to the local user. This becomes critical once authentication is added.

dokipen:
## Summary
- Shell-escape each `ExtraArg` with single quotes before joining into the command template, preventing shell metacharacters from being interpreted when tmux `send-keys` types the rendered command into the pane
- Add `shellEscapeArg` and `shellJoinArgs` helper functions using POSIX single-quote escaping
- Add unit tests covering injection payloads (semicolons, backticks, `$()`, pipes, etc.)
- Add E2E test that verifies a `; touch <file>` injection payload does NOT create the marker file

Fixes #32

## Test plan
- [x] Unit tests for `shellEscapeArg` and `shellJoinArgs` (12 + 6 cases)
- [x] E2E test `TestCreateSession_ExtraArgsInjection` verifies injection is neutralized
- [x] Full test suite passes (`go test ./...`)

### DIFF
--- a/services/agents/internal/session/manager.go
+++ b/services/agents/internal/session/manager.go
@@ -44,8 +44,26 @@ type CreateRequest struct {
 	ExtraArgs    []string
 }
 
+const (
+	maxExtraArgs    = 64
+	maxExtraArgLen  = 4096
+)
+
 // Create validates inputs, creates tmux session, starts command, returns Session.
 func (m *Manager) Create(req CreateRequest) (*Session, error) {
+	// Validate ExtraArgs limits and content.
+	if len(req.ExtraArgs) > maxExtraArgs {
+		return nil, &Error{Code: ErrInvalidArgument, Message: fmt.Sprintf("too many extra_args: %d (max %d)", len(req.ExtraArgs), maxExtraArgs)}
+	}
+	for i, arg := range req.ExtraArgs {
+		if len(arg) > maxExtraArgLen {
+			return nil, &Error{Code: ErrInvalidArgument, Message: fmt.Sprintf("extra_args[%d] too long: %d bytes (max %d)", i, len(arg), maxExtraArgLen)}
+		}
+		if strings.ContainsRune(arg, '\x00') {
+			return nil, &Error{Code: ErrInvalidArgument, Message: fmt.Sprintf("extra_args[%d] contains null byte", i)}
+		}
+	}
+
 	// Validate profile exists.
 	profile, ok := m.profiles[req.AgentProfile]
 	if !ok {
--- a/services/agents/test/e2e/extra_args_injection_test.go
+++ b/services/agents/test/e2e/extra_args_injection_test.go
@@ -8,7 +8,26 @@ import (
 	"time"
 
 	agentsv1 "github.com/dokipen/claude-cadence/services/agents/gen/agents/v1"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
+
+func TestCreateSession_ExtraArgsNullByte(t *testing.T) {
+	client := newTestClient(t)
+	ctx := context.Background()
+
+	_, err := client.CreateSession(ctx, &agentsv1.CreateSessionRequest{
+		AgentProfile: "echo-args",
+		SessionName:  uniqueSessionName(t),
+		ExtraArgs:    []string{"hello\x00world"},
+	})
+	if err == nil {
+		t.Fatal("expected error for null byte in ExtraArgs, got nil")
+	}
+	if code := status.Code(err); code != codes.InvalidArgument {
+		t.Errorf("expected InvalidArgument, got %v", code)
+	}
+}
 
 func TestCreateSession_ExtraArgsInjection(t *testing.T) {
 	client := newTestClient(t)


dokipen (commented):
## Code Review — PR #35: Sanitize ExtraArgs to prevent shell injection

All unit tests pass (`go test ./internal/session/...`), `go vet` and `go build` are clean.

---

### Critical

None.

---

### Warnings

**`templateData.ExtraArgs` field alignment (cosmetic but inconsistent)** — `/services/agents/internal/session/manager.go` lines 231-236

The struct literal has misaligned field tags:
```go
type templateData struct {
    SessionID   string
    SessionName string
    ExtraArgs   string
    WorktreePath string   // no alignment with the others
}
```
`WorktreePath` is missing a leading space to align the type column with `ExtraArgs`. This is a minor gofmt issue; `gofmt` would normalize it. Not a blocker but worth fixing for consistency.

**E2E test uses a hard `time.Sleep`** — `/services/agents/test/e2e/extra_args_injection_test.go` line 48

```go
time.Sleep(1 * time.Second)
```
A fixed 1-second sleep is fragile — on a heavily loaded CI host the `echo` command may not have completed yet, causing a false-negative. The existing E2E helpers (e.g. the session lifecycle test) likely poll for state; the same pattern should be used here. If a polling helper does not yet exist this is worth a follow-up ticket, but it would tighten this test's reliability.

---

### Suggestions

**`shellEscapeArg` and `shellJoinArgs` are unexported** — this is correct for now, but worth a comment that these are intentionally package-private and that callers outside this package should go through `renderCommand`. As the codebase grows a future author might be tempted to duplicate the logic elsewhere.

**No validation on the number or total size of `ExtraArgs`** — a caller can supply an arbitrarily large slice. This is unlikely to matter for an internal gRPC service, but a length cap (e.g., 64 args, 4 KiB total rendered string) would make the attack surface explicit. Consider as a follow-up rather than a blocker.

**`echo-args` profile in `testdata/config.yaml` has no quoting around `{{.ExtraArgs}}`** — line 22:
```yaml
command: "echo {{.ExtraArgs}}"
```
This is intentional for the test (we want to verify the escaping works), but it is also the exact pattern that would be dangerous if `shellJoinArgs` were ever removed. A brief comment in the YAML would document the intent clearly.

---

### Notes

- The POSIX single-quote escaping strategy (`'` -> `'\''`) is correct and is the standard approach. The implementation matches the documented contract.
- `text/template` (not `html/template`) is used, which is correct here — `html/template` would double-escape the already-escaped output.
- The `envKeyRe` allowlist for environment variable keys is a good precedent for the kind of defence-in-depth this PR extends to `ExtraArgs`.

---

**REVIEW STATUS: APPROVED**

The core fix is correct, well-tested, and safe to merge. The two warnings (struct alignment and sleep-based E2E assertion) are polish items that can be addressed in follow-up; neither is a security or correctness risk.

dokipen (commented):
## Security Review: ExtraArgs Shell Injection Sanitization

Overall assessment: the core fix is sound and uses the correct POSIX escaping technique. A few issues need attention before merge, plus several informational notes.

---

### Findings

**Warning: No length limit on ExtraArgs allows resource exhaustion**

`/Users/bob/src/claude-cadence/.worktrees/32-sanitize-extra-args/services/agents/internal/session/manager.go` — `Create()`, lines 48–138

There is no upper bound on the number of ExtraArgs elements or the total length of any individual arg. A caller can send args with megabytes of data that get concatenated into `cmdStr` and passed to tmux. This can exhaust memory in the process and/or hit `tmux send-keys` argument-size limits (OS `ARG_MAX`). Add a maximum per-arg length (e.g., 4096 bytes) and a maximum total arg count (e.g., 64) with an `ErrInvalidArgument` response.

---

**Warning: Null bytes are not rejected and have undefined POSIX shell behavior**

`/Users/bob/src/claude-cadence/.worktrees/32-sanitize-extra-args/services/agents/internal/session/manager.go` — `shellEscapeArg()`, line 261

A Go string (and therefore a protobuf `string` field) can contain embedded null bytes (`\x00`). `strings.ReplaceAll` passes them through unchanged. POSIX defines shell behavior with null bytes inside single-quoted strings as undefined; in practice, bash and dash truncate the argument at the first null, which can cause the remainder of the arg (including the closing `'`) to be misinterpreted, potentially breaking out of the quoted context in some shell implementations. Reject any ExtraArg containing a null byte with `ErrInvalidArgument` before escaping.

---

**Suggestion: E2E test relies on a fixed 1-second sleep — fragile on slow CI**

`/Users/bob/src/claude-cadence/.worktrees/32-sanitize-extra-args/services/agents/test/e2e/extra_args_injection_test.go`, line 48

`time.Sleep(1 * time.Second)` is a fixed wait for the tmux pane to execute. On a heavily loaded CI runner this is prone to false-pass (injection ran but test checked before the file appeared) or unnecessary slowness. Consider polling with a short deadline (e.g., 100 ms intervals up to 3 s) or, better, have the `echo-args` profile write to a known output file that the test reads to confirm execution completed before checking for the marker.

---

**Suggestion: E2E injection test covers only one payload vector**

`/Users/bob/src/claude-cadence/.worktrees/32-sanitize-extra-args/services/agents/test/e2e/extra_args_injection_test.go`, lines 24–53

The test validates `; touch <file>` but not other injection patterns that behave differently at the shell boundary: newline-terminated commands (`\ntouch <file>`), backtick substitution (`` `touch <file>` ``), or `$(touch <file>)`. Unit tests cover these in isolation, but an E2E test for at least the newline case is valuable because the rendered string is passed as a single `send-keys` argument — tmux may interpret embedded newlines as Enter, which the unit-level escaping does not address.

Specifically: `shellEscapeArg("foo\ntouch " + markerFile)` produces `'foo\ntouch /path'` which a POSIX shell will treat as a literal newline inside the string — safe. But tmux's `send-keys` with a literal newline character in the key string may interpret it as pressing Enter mid-sequence. This scenario is not covered by the current E2E test.

---

**Suggestion: Add a `renderCommand` integration unit test**

`/Users/bob/src/claude-cadence/.worktrees/32-sanitize-extra-args/services/agents/internal/session/manager_test.go`

The unit tests validate `shellEscapeArg` and `shellJoinArgs` in isolation, but there is no test that exercises `renderCommand` end-to-end (template parsing + data substitution + ExtraArgs embedding). A table-driven test verifying that `renderCommand` with a template like `"claude {{.ExtraArgs}}"` and an injection payload produces the expected escaped output would close the gap between the escaping helpers and their point of use.

---

**Note: `SetEnv` value is safe — no escaping needed**

`/Users/bob/src/claude-cadence/.worktrees/32-sanitize-extra-args/services/agents/internal/tmux/tmux.go`, line 54

`SetEnv` passes the value as a discrete `exec.Command` argument, which goes through the OS `execve` argument vector with no shell involved. The value never touches a shell, so no escaping is required. The existing key-format validation via `envKeyRe` is sufficient.

---

**Note: POSIX single-quote escaping technique is correct**

`/Users/bob/src/claude-cadence/.worktrees/32-sanitize-extra-args/services/agents/internal/session/manager.go`, lines 261–263

The `'` → `'\''` substitution is the standard, correct POSIX approach. Inside single quotes no characters have special meaning except `'` itself, so this technique neutralizes all shell metacharacters (`$`, `` ` ``, `\`, `!`, `;`, `|`, `&`, `(`, `)`, `{`, `}`, `<`, `>`, `*`, `?`, `~`, space, tab) in a single pass without character enumeration. The implementation is correct.

---

### Summary

| Finding | Severity | Location |
|---------|----------|----------|
| No length/count limit on ExtraArgs | Warning | `manager.go` `Create()` |
| Null bytes bypass escaping correctness | Warning | `manager.go` `shellEscapeArg()` |
| Fixed-sleep E2E timing | Suggestion | `extra_args_injection_test.go:48` |
| E2E test covers only one injection vector | Suggestion | `extra_args_injection_test.go` |
| Missing `renderCommand` integration test | Suggestion | `manager_test.go` |
| `SetEnv` value path is safe | Note | `tmux.go:54` |
| POSIX escaping technique is correct | Note | `manager.go:261–263` |

The two Warnings should be resolved before merge. The null-byte issue in particular is low-effort to fix (`strings.ContainsRune(arg, 0)` check before escaping) and closes a real, if obscure, correctness gap.

PR #36, opened 2026-03-15
Add comment CRUD resolvers, CLI commands, and E2E tests (#8)

### LINKED ISSUE #8: Comment CRUD + CLI + E2E
## Description

Implement comment CRUD (add, view on ticket, edit own, delete own) as GraphQL resolvers, CLI commands, and E2E tests.

### GraphQL Operations
- `addComment(ticketId, body)` mutation
- `updateComment(id, body)` mutation
- `deleteComment(id)` mutation
- Comments included in ticket query responses with author info

### CLI Commands
- `issues comment add <ticket-id> --body "..."`
- `issues comment edit <id> --body "..."`
- `issues comment delete <id>`

## Acceptance Criteria

- [ ] `addComment`, `updateComment`, `deleteComment` mutations work
- [ ] Comments include author info in responses
- [ ] CLI commands: `issues comment add/edit/delete`
- [ ] E2E tests pass: comments.e2e.ts

dokipen:
## Summary
- Add `addComment`, `updateComment`, `deleteComment` GraphQL mutations with validation
- Add `Comment.author` field resolver for direct mutation responses
- Add CLI commands: `comment add`, `comment edit`, `comment delete`
- Add 7 E2E tests covering full comment lifecycle (add, view, edit, delete)

Fixes #8

## Test plan
- [x] All 7 new comment E2E tests pass
- [x] All 64 tests across 7 test files pass (no regressions)

### DIFF
--- a/services/issues-cli/src/commands/comment.ts
+++ b/services/issues-cli/src/commands/comment.ts
@@ -102,7 +102,7 @@ export function registerCommentCommand(program: Command): void {
         spinner.succeed("Comment updated");
         const c = data.updateComment;
         console.log();
-        console.log(`  ${chalk.bold(`@${c.author.login}`)}  ${chalk.dim(c.createdAt)}`);
+        console.log(`  ${chalk.bold(`@${c.author.login}`)}  ${chalk.dim(c.updatedAt)}`);
         console.log(`    ${c.body}`);
         console.log();
       } catch (error) {
--- a/services/issues/src/schema/resolvers/ticket.ts
+++ b/services/issues/src/schema/resolvers/ticket.ts
@@ -340,6 +340,8 @@ export const ticketResolvers = {
       { ticketId, body }: { ticketId: string; body: string },
       { prisma }: Context
     ) => {
+      if (!body.trim()) throw new Error("Comment body cannot be empty");
+
       const ticket = await prisma.ticket.findUnique({ where: { id: ticketId } });
       if (!ticket) throw new Error(`Ticket not found: ${ticketId}`);
 
@@ -356,6 +358,8 @@ export const ticketResolvers = {
       { id, body }: { id: string; body: string },
       { prisma }: Context
     ) => {
+      if (!body.trim()) throw new Error("Comment body cannot be empty");
+
       const existing = await prisma.comment.findUnique({ where: { id } });
       if (!existing) throw new Error("Comment not found");
 


dokipen (commented):
## Code Review: Comment CRUD (PR #36)

Both services build cleanly with `tsc`. The implementation is consistent with existing patterns. Here are my findings:

---

**Critical** (must fix):

- **`addComment` uses `findFirstOrThrow` to hard-code the author** (`services/issues/src/schema/resolvers/ticket.ts`, line 346). The resolver calls `prisma.user.findFirstOrThrow()` to pick an arbitrary first user from the database as the comment author. There is no authenticated user concept yet, but this is worse than a stub — it silently assigns ownership to a random user, which will produce incorrect data in multi-user scenarios and is non-obvious behavior. The same pattern does not exist anywhere else in the codebase. At minimum this should be a hard-coded sentinel value, an explicit TODO comment explaining the limitation, or should require an `authorId` argument until auth is wired up. As written, it is a latent data-integrity bug.

---

**Warnings** (should fix):

- **`Comment.author` field resolver reuses `assigneeByUserId` loader** (`resolvers/ticket.ts`, line 387). The loader is semantically named for ticket assignees but is being pressed into service for comment authors. The loader itself is correct (it loads users by ID), but the naming is misleading and will cause confusion when reading the code. Consider renaming to `userById` or adding a dedicated `authorByUserId` alias in the loaders interface to make intent clear.

- **`deleteComment` fetches the comment with `include: { author: true }` but `Comment.author` has a field resolver** (`resolvers/ticket.ts`, lines 374-376). The `deleteComment` resolver explicitly includes `author` in the Prisma query so the returned object has the author populated. However the `Comment.author` field resolver (line 386-388) will also fire on the returned object and call `loaders.assigneeByUserId.load(parent.authorId)`. This works today because DataLoader deduplicates, but the explicit `include` is redundant and inconsistent with how `updateComment` (line 362-366) handles the same return type. Both mutations should use the same approach — rely on the field resolver, not the eager include.

- **Missing test for error cases** (`test/e2e/comments.e2e.ts`). The seven tests cover the happy path thoroughly. None of the other E2E test files test error cases either, so this is consistent with the pattern, but the `deleteComment` double-delete scenario and `updateComment` on a non-existent ID are particularly useful to catch regressions in the "not found" guard paths added in this PR.

---

**Suggestions** (nice to have):

- **`comment edit` output shows `createdAt` instead of `updatedAt`** (`commands/comment.ts`, line 105). The `edit` command fetches both `createdAt` and `updatedAt` from the server but displays `createdAt` in the timestamp position. For an edit confirmation, showing `updatedAt` is more informative.

- **`deleteComment` output is minimal** (`commands/comment.ts`, lines 127-130). The delete confirmation prints only the truncated comment body. Other destructive confirmations in the codebase (e.g., `block remove`) show the affected entity title. Showing the ticket context (even just the `ticketId`) would help the user confirm they deleted the right comment.

- **`beforeAll` in `comments.e2e.ts` does not assert user creation** (line 14-23). The `createUser` mutation result is not captured or checked. If user creation silently fails, the subsequent `addComment` calls would use whatever user happens to already exist in the DB (due to the `findFirstOrThrow` issue above), giving a false-positive test. Asserting the returned `id` would make the test setup self-verifying.

---

REVIEW STATUS: CHANGES NEEDED
- Fix `addComment` author selection (critical data integrity issue)
- Rename or alias `assigneeByUserId` loader for comment author use
- Fix redundant `include: { author: true }` in `deleteComment` resolver

dokipen (commented):
## Security Review: Comment CRUD (PR #36)

### Risk Assessment

| Finding | Severity | Location | Recommendation |
|---------|----------|----------|----------------|
| Broken author attribution — `findFirstOrThrow()` picks an arbitrary user as comment author | High | `ticket.ts:346` | Pass authenticated caller identity via context; do not infer author from DB |
| No authorization on `updateComment` / `deleteComment` — any caller can mutate any comment | High | `ticket.ts:354-382` | Enforce ownership check: `existing.authorId === ctx.currentUser.id` before write |
| No authorization on `deleteComment` — any caller can delete any comment | High | `ticket.ts:369-382` | Same as above |
| No body length / content validation on `addComment` / `updateComment` | Medium | `ticket.ts:338-367` | Enforce a maximum body length (e.g. 10,000 chars) and reject empty strings |
| No authentication layer on the GraphQL server — all mutations are unauthenticated | Medium | `src/index.ts:20-27` | Add an auth middleware or token-based context; this affects all mutations, not just comments |
| `assignTicket` does not verify that `userId` exists before writing | Low | `ticket.ts:232-241` | Add a `findUnique` guard for `userId` as done in `addLabel` |
| Deleted comment body echoed back in CLI output without sanitization | Low | `comment.ts:129` | Acceptable in a terminal context; note for completeness |

---

### Detailed Findings

#### [High] Arbitrary user selected as comment author

In `addComment`, the author is determined by:

```ts
const user = await prisma.user.findFirstOrThrow();
```

`findFirst` with no filter returns whichever row the database chooses. This means:
- Every comment is attributed to the same (arbitrary) user.
- When a real auth layer is added later, existing comment records will have the wrong `authorId`.
- An attacker who can influence ordering or insert a user record could cause comments to be attributed to a different account.

**Recommendation:** The caller's identity must be resolved from the request context (e.g. a JWT or session token decoded in the Apollo `context` function) and stored as `ctx.currentUser`. Pass that ID as `authorId`.

---

#### [High] Missing ownership checks on `updateComment` and `deleteComment`

Neither resolver verifies that the caller owns the comment being modified:

```ts
// updateComment — no ownership check
const existing = await prisma.comment.findUnique({ where: { id } });
if (!existing) throw new Error("Comment not found");
return prisma.comment.update({ ... });

// deleteComment — same pattern
```

Because the GraphQL server has no authentication middleware (`src/index.ts` sets up an empty context), any unauthenticated client can edit or delete any comment by ID. This is a BOLA/IDOR vulnerability.

**Recommendation:** Once auth context is added, assert `existing.authorId === ctx.currentUser.id` (or a moderator role) before proceeding. Raise a generic "Not found" error rather than "Unauthorized" to avoid leaking existence.

---

#### [Medium] No input validation on comment body

`addComment` and `updateComment` accept `body: String!` with no server-side constraints. The GraphQL schema enforces non-null but not:
- Minimum length (empty string `""` is accepted)
- Maximum length (arbitrarily large payloads go directly to the DB)

**Recommendation:** Add a guard at the top of each resolver:

```ts
if (!body.trim().length) throw new Error("Comment body cannot be empty");
if (body.length > 10_000) throw new Error("Comment body exceeds maximum length");
```

---

#### [Medium] No authentication on the GraphQL server

`src/index.ts` creates the Apollo context with only `prisma` and `loaders` — no user identity, no token validation. All mutations (create/update/delete ticket, assign, transition state, add/edit/delete comment) are open to any process that can reach the port.

This is a pre-existing gap, but the comment CRUD mutations increase the attack surface. **Recommendation:** Add token verification in the `context` callback before merging further mutation work.

---

### Injection Risk Assessment

**No injection risk found.** All database access goes through Prisma's parameterised query builder. The CLI passes arguments as GraphQL variables (not string interpolation into query documents), so there is no SQL injection or GraphQL injection surface.

---

### Dependency Audit

Both `services/issues` and `services/issues-cli` returned clean audits:

```
vulnerabilities: { info: 0, low: 0, moderate: 0, high: 0, critical: 0 }
```

No action required on dependencies.

---

### Summary

**Overall risk: High** — the two high-severity findings (phantom author attribution and missing ownership checks) are architectural issues that will become harder to retrofit as the feature grows. They are not exploitable in the current single-user dev environment, but they establish patterns that will be dangerous if the service is ever exposed to multiple users or the network.

**Prioritized remediation:**
1. Add an authenticated `currentUser` to the Apollo request context (`src/index.ts`).
2. Use `ctx.currentUser.id` as `authorId` in `addComment` — remove the `findFirstOrThrow` call.
3. Add ownership assertions to `updateComment` and `deleteComment`.
4. Add body length validation to both comment write resolvers.
5. Add a `findUnique` guard for `userId` in `assignTicket` (low priority, pre-existing gap).

PR #38, opened 2026-03-15
Add gRPC token authentication and configurable reflection

### LINKED ISSUE #31: Agent service: add gRPC authentication and TLS
## Description
Discovered during PR #29 code review. The gRPC server has no authentication or TLS, and reflection is enabled unconditionally.

## Acceptance Criteria
- [ ] Add TLS support (at minimum for non-localhost bindings)
- [ ] Add authentication mechanism (mTLS or token-based)
- [ ] Make gRPC reflection configurable (disable by default in production)

## Notes
Phase 1 is local-only by design (127.0.0.1), so this is acceptable for now but must be addressed before any network-facing deployment.

dokipen:
## Summary
- Add token-based authentication via bearer token unary/stream interceptors
- Make gRPC reflection opt-in (disabled by default) via `reflection` config field
- Config validation rejects token mode without a token value and invalid auth modes
- All changes are backward-compatible — defaults preserve existing behavior (auth: none, reflection: off)

**Note:** TLS is intentionally excluded — TLS termination will be handled by a reverse proxy.

## Files Changed
- `internal/config/config.go` — AuthConfig struct, validation, ResolveToken
- `internal/server/server.go` — conditional interceptors, conditional reflection
- `internal/server/auth.go` — bearer token validation interceptors (new)
- `cmd/agentd/main.go` — pass config to server.New
- `test/e2e/helpers_test.go` — use new server.New signature

## Test plan
- [x] 10 config validation unit tests (valid/invalid auth combos, ResolveToken)
- [x] 6 auth interceptor unit tests (valid/missing/wrong token)
- [x] 2 reflection registration tests (enabled/disabled)
- [ ] E2E tests still pass (require tmux)

Fixes #31

### DIFF
--- a/services/agents/cmd/agentd/main.go
+++ b/services/agents/cmd/agentd/main.go
@@ -12,6 +12,7 @@ import (
 	"github.com/dokipen/claude-cadence/services/agents/internal/service"
 	"github.com/dokipen/claude-cadence/services/agents/internal/session"
 	"github.com/dokipen/claude-cadence/services/agents/internal/tmux"
+	"github.com/dokipen/claude-cadence/services/agents/internal/ttyd"
 )
 
 func main() {
@@ -57,8 +58,9 @@ func main() {
 
 	// Create components.
 	tmuxClient := tmux.NewClient(cfg.Tmux.SocketName)
+	ttydClient := ttyd.NewClient(cfg.Ttyd.Enabled, cfg.Ttyd.BasePort)
 	store := session.NewStore()
-	manager := session.NewManager(store, tmuxClient, cfg.Profiles)
+	manager := session.NewManager(store, tmuxClient, ttydClient, cfg.Profiles)
 	agentService := service.NewAgentService(manager)
 
 	srv, err := server.New(agentService, cfg)
--- a/services/agents/config.example.yaml
+++ b/services/agents/config.example.yaml
@@ -12,6 +12,11 @@ root_dir: "/var/lib/agentd"
 tmux:
   socket_name: "agentd"
 
+# ttyd websocket terminal
+ttyd:
+  enabled: false
+  base_port: 7681
+
 # Logging
 log:
   level: "info"    # debug, info, warn, error
--- a/services/agents/internal/config/config.go
+++ b/services/agents/internal/config/config.go
@@ -30,6 +30,7 @@ type Config struct {
 	Port       int                `yaml:"port"`
 	RootDir    string             `yaml:"root_dir"`
 	Tmux       TmuxConfig         `yaml:"tmux"`
+	Ttyd       TtydConfig         `yaml:"ttyd"`
 	Log        LogConfig          `yaml:"log"`
 	Profiles   map[string]Profile `yaml:"profiles"`
 	Auth       AuthConfig         `yaml:"auth"`
@@ -39,6 +40,12 @@ type Config struct {
 // TmuxConfig holds tmux-specific settings.
 type TmuxConfig struct {
 	SocketName string `yaml:"socket_name"`
+}
+
+// TtydConfig holds ttyd websocket terminal settings.
+type TtydConfig struct {
+	Enabled  bool `yaml:"enabled"`
+	BasePort int  `yaml:"base_port"`
 }
 
 // LogConfig holds logging settings.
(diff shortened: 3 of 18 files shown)


dokipen (commented):
## Code Review – PR #38: gRPC token auth and configurable reflection

**REVIEW STATUS: CHANGES NEEDED**

---

### Critical

**1. Token comparison is not timing-safe (`services/agents/internal/server/auth.go`, line 46)**

```go
if token[len(prefix):] != expected {
```

Go's `!=` operator on strings is NOT constant-time. It short-circuits on the first differing byte, which allows a timing oracle: an attacker who can make many requests and measure response latency can recover the token one byte at a time.

Fix: use `crypto/subtle.ConstantTimeCompare`:

```go
import "crypto/subtle"

provided := []byte(token[len(prefix):])
secret    := []byte(expected)
if subtle.ConstantTimeCompare(provided, secret) != 1 {
    return status.Error(codes.Unauthenticated, "invalid token")
}
```

The same pattern should be applied to the `Bearer ` prefix check on line 43 if the prefix itself could be considered sensitive (it is not in this case, so the existing slice-length check there is fine).

---

### Warnings

**2. No validation prevents binding to a public interface with `auth.mode: none` (`services/agents/internal/config/config.go`, `validate` function)**

`config_test.go` line 84-93 (`TestValidate_NonLocalhostWithTokenAuth`) shows that `0.0.0.0` is an accepted host for token mode, which is correct. However, there is no corresponding check that rejects (or at least warns about) `auth.mode: none` when the host is not a loopback address.

A config like `host: 0.0.0.0, auth.mode: none` silently exposes an unauthenticated gRPC service to the network. At minimum, `validate` should return an error (or a clearly documented warning path) for that combination. The test `TestValidate_ValidLocalhostNoAuth` passes `127.0.0.1`, which is fine, but nothing prevents a user from setting `host: 0.0.0.0` with no auth.

Suggested addition to `validate`:

```go
if cfg.Auth.Mode == "none" && !isLoopback(cfg.Host) {
    return fmt.Errorf("auth.mode \"none\" is not permitted when host is not a loopback address")
}
```

**3. E2E tests never exercise the token-auth path (`services/agents/test/e2e/helpers_test.go`, lines 38-43)**

`helpers_test.go` hard-codes `auth.mode: none` for the e2e test server. There are no e2e (or even integration) tests that:
- Start a server with `auth.mode: token`
- Confirm that an unauthenticated call is rejected with `codes.Unauthenticated`
- Confirm that a correctly-authenticated call succeeds

The unit tests in `auth_test.go` cover `validateToken` in isolation, but there is no test that verifies the interceptors are actually wired into the server and invoked end-to-end. A server wiring mistake (e.g., wrong interceptor order or the interceptors not being attached) would not be caught.

---

### Suggestions

**4. Duplicate `os.Setenv` call in config test (`services/agents/internal/config/config_test.go`, line 112)**

```go
t.Setenv("TEST_AGENT_TOKEN", "")
_ = os.Setenv("TEST_AGENT_TOKEN", "")  // redundant
```

`t.Setenv` already calls `os.Setenv` and registers cleanup. The second call is dead code and should be removed.

**5. Token value logged nowhere but also not guarded**

`ResolveToken` returns the raw token string. There is no risk of it appearing in logs in the current code, but consider adding a `String() string` method to `AuthConfig` (or a `GoString`) that redacts the `Token` field so that accidental `fmt.Printf("%+v", cfg)` calls in future code do not leak it.

**6. `validate` error message wrapping inconsistency**

`Load` wraps the validation error with `"validating config: %w"`, so callers see `"validating config: token or token_env_var required..."`. The other error paths (`"reading config file: ..."`, `"parsing config file: ..."`) are consistent with each other. This is fine as-is but worth noting that the test at `config_test.go` line 66-69 asserts the unwrapped message directly against `err.Error()` while bypassing `Load`, so the assertion would silently pass even if `Load`'s wrapping changed.

---

### Summary

| # | Severity | File | Line |
|---|----------|------|------|
| 1 | Critical | `internal/server/auth.go` | 46 |
| 2 | Warning  | `internal/config/config.go` | `validate()` |
| 3 | Warning  | `test/e2e/helpers_test.go` | 38-43 |
| 4 | Suggestion | `internal/config/config_test.go` | 112 |
| 5 | Suggestion | `internal/config/config.go` | `AuthConfig` |
| 6 | Suggestion | `internal/config/config_test.go` | 66-69 |

Items 1–3 must be resolved before merge.

PR #53, opened 2026-03-15
Add install/uninstall scripts and service templates

dokipen:
## Summary
- Interactive `install.sh` installer: detects macOS vs Linux, prompts for user/host/port/directories, verifies prerequisites (git, tmux), generates config, installs binary, renders service templates, starts service with health check
- `uninstall.sh`: stops service, removes binary and service definition, optionally removes data directories
- `agentd.plist.tmpl`: launchd template with KeepAlive, RunAtLoad, configurable log paths
- `agentd.service.tmpl`: systemd unit with Restart=on-failure, journal logging
- Fixes pre-existing `server.New` signature mismatch in `git_worktree_test.go`

Fixes #20

## Test plan
- [x] 4 new e2e tests: PlistGeneration, SystemdGeneration, PlistCustomValues, SystemdCustomValues
- [x] Full e2e test suite passes (24 pass, 4 skip)
- [x] Unit tests pass
- [x] shellcheck clean on both scripts
- [ ] Manual install test on macOS
- [ ] Manual install test on Linux


dokipen (commented):
## Code Review: Add install/uninstall scripts and service templates

All acceptance criteria from issue #20 are met. shellcheck passes clean on both scripts, all 4 e2e tests pass. The structure and overall approach are solid. Several issues need attention before merge.

---

### Findings

#### Critical

**Shell injection via `eval` in `prompt()` — `install.sh` line 27**

```bash
eval "$var_name=\"${value:-$default}\""
```

`value` is raw user input fed directly into `eval`. A user who enters `$(rm -rf /)` or `\"; malicious_cmd; echo \"` as a directory path causes immediate command execution. `var_name` is also sourced from the calling context; if this function were ever exposed to external input the injection extends there too.

Fix: use `printf -v` instead of `eval`:
```bash
printf -v "$var_name" '%s' "${value:-$default}"
```
`printf -v` assigns to a named variable without invoking a shell interpreter on the value.

---

#### Warnings

**`render_template` sed substitutions break on `|` in paths — `install.sh` lines 184-191**

The sed delimiter is `|`, so any variable containing a literal `|` (unusual but valid in paths) silently corrupts the output or produces a sed error. A safe pattern is to pre-escape the delimiter in each variable, or use a tool that does not have delimiter collisions (e.g. `awk` with `gsub`). At minimum a comment warning about this constraint would help operators.

**macOS user creation is incomplete — `install.sh` lines 77-80**

The `dscl` block does not set `UniqueID` or `PrimaryGroupID`, which are required fields for a valid macOS user. Without them `id "$AGENTD_USER"` will still succeed (the directory entry exists) but the account is unusable by the OS for running processes. The subsequent `launchctl bootstrap` will fail or run the daemon as nobody/root depending on macOS version.

Minimum required fields:
```bash
sudo dscl . -create "/Users/$AGENTD_USER" UniqueID   <unused-uid>
sudo dscl . -create "/Users/$AGENTD_USER" PrimaryGroupID 20
```
It's worth noting that `sysadminctl -addUser` is the recommended approach on modern macOS (10.13+) and handles all required fields atomically.

**Config file write fails for non-root on Linux — `install.sh` lines 114, 132**

`setup_directories` creates `AGENTD_CONFIG_DIR` with `mkdir -p "$AGENTD_CONFIG_DIR"` (no `sudo`), and `generate_config` writes to it with a plain `cat >`. On Linux the default config dir is `/etc/agentd`, which requires root. Both calls need `sudo` on Linux, or the script should detect whether the target path requires elevated permissions and act accordingly.

**`health_check` failure aborts installation — `install.sh` line 291**

`health_check` returns 1 on timeout and `main` calls it without `|| true`. With `set -e` in effect, a slow-starting service will cause the script to exit with a non-zero code after the service has already been installed, leaving the user confused about whether installation succeeded. Consider either calling `health_check || true` (since it already emits a warning) or restructuring so the installer prints "installation complete" regardless of health-check outcome.

---

#### Suggestions

**`service_dir` computed via unresolved `..` — `install.sh` line 163**

```bash
local service_dir="$script_dir/.."
```

This works when the script is run directly but breaks if the script is symlinked (the `..` resolves relative to the symlink target's parent, not the binary's install tree). Use `realpath` or `dirname` on `script_dir` to normalize:
```bash
local service_dir
service_dir="$(dirname "$script_dir")"
```

**Test coverage gap: special characters in paths**

The e2e tests verify that all placeholders are replaced and the correct values appear. They do not test the case where a substituted value contains the sed delimiter (`|`) or other sed metacharacters. Adding a test case with a path like `/opt/pipe|test/agentd` would catch the delimiter-collision bug described above.

**`uninstall.sh` does not prompt for config directory**

`remove_data` hard-codes a fixed list of candidate directories (`/var/lib/agentd`, `$HOME/.config/agentd`, etc.). If the user installed to a custom config path during `install.sh`, that path will not be cleaned up. Consider reading the config path from the installed plist or unit file before removing them, or prompting the user similarly to how `install.sh` does.

---

### Verdict

REVIEW STATUS: CHANGES NEEDED
- [Critical] Shell injection via `eval` in `prompt()` — replace with `printf -v`
- [Warning] Incomplete macOS user creation via `dscl` — missing `UniqueID`/`PrimaryGroupID`
- [Warning] Config directory write on Linux requires sudo
- [Warning] `health_check` failure exits the installer with non-zero after successful installation

PR #54, opened 2026-03-15
Add Vault integration for secrets and private repo credentials

### LINKED ISSUE #18: Agent service: Vault integration for secrets
## Description

Integrate HashiCorp Vault for fetching credentials needed to clone private repositories and for injecting secrets as environment variables into agent sessions.

See `services/agents/docs/PLAN.md` (Phase 3) for full details.

## What Gets Built

- `internal/vault/vault.go` — Vault client (token + AppRole auth), secret reading
- Update `internal/git/git.go` — accept credentials for private repo clone/fetch
- Update `internal/session/manager.go` — fetch vault secrets, inject as env vars
- `test/e2e/vault_test.go` — tests with `vault server -dev`

## Acceptance Criteria

- [ ] Profiles with `vault_secret` get credentials injected for git operations
- [ ] Vault secrets injected as env vars into tmux sessions
- [ ] Token auth method works
- [ ] AppRole auth method works
- [ ] Profiles without `vault_secret` work without Vault (public repos)
- [ ] E2e tests pass with Vault dev server

## Notes

- User story: `services/agents/docs/user-stories/03-vault-secrets.md`
- Blocked by #17 (git + worktrees)

dokipen:
## Summary

Fixes #18

- Add `VaultConfig` and `Profile.VaultSecret` to config types with validation
- Create `internal/vault/vault.go` — Vault client supporting token and AppRole authentication
- Update `git.go` to accept optional credentials for HTTPS (token via credential helper) and SSH (key via `GIT_SSH_COMMAND`) repos
- Update session manager to fetch Vault secrets before clone and inject them as environment variables into tmux sessions
- Wire Vault client initialization in `main.go`
- Add 4 E2E tests: `TestVault_TokenAuth`, `TestVault_EnvInjection`, `TestVault_NoSecret_PublicRepo`, `TestVault_SecretRetrieval`

Also fixes a pre-existing `server.New` signature mismatch in `git_worktree_test.go`.

## Test plan

- [x] All existing tests pass (config, server, session, e2e)
- [x] `TestVault_TokenAuth` — Vault token auth works for session creation
- [x] `TestVault_EnvInjection` — Vault secrets injected as env vars in tmux
- [x] `TestVault_NoSecret_PublicRepo` — Profiles without vault_secret work fine
- [x] `TestVault_SecretRetrieval` — Secrets fetched and available in session

### DIFF
--- a/services/agents/internal/git/git.go
+++ b/services/agents/internal/git/git.go
@@ -86,7 +86,8 @@ func (c *Client) EnsureClone(repoURL string, creds *Credentials) (string, error)
 	}
 
 	cmd := exec.Command("git", "clone", repoURL, cloneDir)
-	c.applyCredentials(cmd, repoURL, creds)
+	cleanup := c.applyCredentials(cmd, repoURL, creds)
+	defer cleanup()
 	if output, err := cmd.CombinedOutput(); err != nil {
 		return "", fmt.Errorf("git clone: %w: %s", err, string(output))
 	}
@@ -193,10 +194,15 @@ func (c *Client) pullDefaultBranch(cloneDir string, creds *Credentials) error {
 	fetch := exec.Command("git", "-C", cloneDir, "fetch", "origin", branch)
 	// Determine repo URL for credential type detection.
 	urlCmd := exec.Command("git", "-C", cloneDir, "remote", "get-url", "origin")
+	var cleanup func()
 	if urlOut, err := urlCmd.Output(); err == nil {
-		c.applyCredentials(fetch, strings.TrimSpace(string(urlOut)), creds)
+		cleanup = c.applyCredentials(fetch, strings.TrimSpace(string(urlOut)), creds)
+	} else {
+		cleanup = func() {}
 	}
-	if output, err := fetch.CombinedOutput(); err != nil {
+	output, err := fetch.CombinedOutput()
+	cleanup()
+	if err != nil {
 		return fmt.Errorf("git fetch: %w: %s", err, string(output))
 	}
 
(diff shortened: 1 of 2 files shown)


dokipen (commented):
## Code Review — PR #54: Vault Integration

Build and `go vet` pass cleanly. The overall design is sound: config validation is thorough, the credential injection approach uses environment variables rather than modifying repo URLs (which avoids leaking secrets in git reflog/logs), and the E2E tests cover the main paths well.

Below are the issues I found, organized by priority.

---

### Critical

**1. SSH key temp file is never cleaned up — credential leak**
`services/agents/internal/git/git.go` — `writeSSHKey` / `applyCredentials`

`writeSSHKey` creates a temp file and returns its path, but nothing ever removes it. Every git clone or fetch over SSH leaks a PEM private key on the filesystem until the OS cleans up `/tmp`. The fix is to register cleanup with the `exec.Cmd` or, more reliably, use `t.TempDir()`-style cleanup tied to the operation lifetime.

Minimal fix pattern:
```go
keyFile, err := writeSSHKey(creds.SSHKey)
if err == nil {
    defer os.Remove(keyFile)   // <— add this
    cmd.Env = append(cmd.Env, ...)
}
```
Because `cmd.Env` is set before the command runs and the key file only needs to exist for the duration of `cmd.Run()`, a `defer` immediately after creation is safe.

**2. HTTPS credential helper injects token into a shell function — shell injection risk**
`services/agents/internal/git/git.go`, line ~233

```go
helper := fmt.Sprintf("!f() { echo \"password=%s\"; echo \"username=x-access-token\"; }; f", creds.Token)
```

The token is interpolated directly into a shell function string. A token containing `"`, `;`, `$`, or backticks would break out of the shell context. Tokens from Vault should be treated as untrusted strings here.

The standard approach is to write a tiny credential helper script to disk (similar to what is already done for SSH keys), or to use `git credential-store` with a temp file, which accepts arbitrary bytes safely. Using `GIT_ASKPASS` pointing to a script that just echoes the token is another safe pattern.

---

### Warnings

**3. `http.DefaultClient` is shared and has no timeout**
`services/agents/internal/vault/vault.go`, line 24

```go
httpClient: http.DefaultClient,
```

`http.DefaultClient` has no timeout. A hung or slow Vault server will block a session creation call indefinitely. The client should use a private `*http.Client` with a reasonable timeout (e.g., 10–30 seconds):

```go
httpClient: &http.Client{Timeout: 15 * time.Second},
```

Mutating `http.DefaultClient` from outside the package is also always a latent risk when a shared client is stored.

**4. KV v1 / v2 format detection is ambiguous for nested data**
`services/agents/internal/vault/vault.go`, lines 81–86

The KV v2 detection heuristic attempts to unmarshal `data` as `{ "data": map }`, then falls back to v1. A KV v1 secret that _happens_ to have a top-level key named `"data"` whose value is an object will be misidentified as KV v2, silently returning only the nested sub-map. This is an edge case but is hard to debug.

The conventional approach is to require the operator to specify the KV version explicitly (e.g., `kv_version: 1` or `kv_version: 2` in `VaultConfig`), or to require KV v2 paths (which always contain `/data/`) and fail clearly for v1.

**5. AppRole `auth_method` validation does not require `SecretID` or `SecretIDEnvVar`**
`services/agents/internal/config/config.go`, lines 185–188

The `validate` function checks `RoleID` is present for AppRole but does not verify that either `SecretID` or `SecretIDEnvVar` is set. A misconfigured deployment will pass validation then fail at runtime with a less helpful error. Add:

```go
case "approle":
    if cfg.Vault.RoleID == "" {
        return fmt.Errorf("vault.role_id is required for approle auth")
    }
    if cfg.Vault.SecretID == "" && cfg.Vault.SecretIDEnvVar == "" {
        return fmt.Errorf("vault.secret_id or vault.secret_id_env_var required for approle auth")
    }
```

**6. Vault secrets logged indirectly via `fmt.Sprintf("%v", v)` with `slog.Warn`**
`services/agents/internal/session/manager.go`, line 224

The key name is logged on the warning path — that's fine — but `strVal` is the full secret value. The warning at line 225 logs the `envKey` but not the value, which is correct. However, if `fmt.Sprintf("%v", v)` is ever changed to include the value in a debug statement during future iteration, a secret ends up in logs. Consider using a typed `string` assertion rather than `fmt.Sprintf` to make the intent explicit and reduce future accident risk:

```go
strVal, ok := v.(string)
if !ok {
    slog.Warn("skipping vault secret with non-string value", "key", envKey)
    continue
}
```

This also avoids injecting `"<nil>"`, `"true"`, numeric representations, etc. as environment variable values for non-string secrets.

---

### Suggestions

**7. `gitCreds` always allocated even when both fields are empty**
`services/agents/internal/session/manager.go`, lines 150–156

```go
gitCreds = &git.Credentials{}
if token, ok := secrets["token"].(string); ok { gitCreds.Token = token }
if sshKey, ok := secrets["ssh_key"].(string); ok { gitCreds.SSHKey = sshKey }
```

If neither `"token"` nor `"ssh_key"` is in the Vault secret, `gitCreds` is a non-nil struct with empty fields. `applyCredentials` will set `GIT_TERMINAL_PROMPT=0` but apply no actual credentials — which is harmless but surprising. Consider leaving `gitCreds` nil if no credentials were extracted, or documenting this behaviour.

**8. `time.Sleep` in E2E tests**
`services/agents/test/e2e/vault_test.go`, lines ~270, ~333

`time.Sleep(500 * time.Millisecond)` is a fragile way to wait for tmux env vars to be set. This is a minor flakiness risk under load. A short polling loop with `exec.Command("tmux", "show-environment", ...)` until success or timeout would be more robust, consistent with how other E2E tests in the codebase handle similar timing.

---

### Summary

Two issues need fixing before merge: the SSH key temp file leak (Critical #1) and the shell injection vector in the HTTPS credential helper (Critical #2). The HTTP timeout (Warning #3) and AppRole secret_id validation (Warning #5) are straightforward additions. The rest are lower-priority but worth addressing.

REVIEW STATUS: CHANGES NEEDED
- [Critical] SSH temp key file never deleted — PEM key leak on disk
- [Critical] HTTPS credential helper injects token into shell string without escaping — injection risk
- [Warning] `http.DefaultClient` used without timeout in Vault HTTP client
- [Warning] AppRole config validation does not require `secret_id`/`secret_id_env_var`

dokipen:
## Review Findings Addressed

All critical and warning findings from code review and security review have been fixed in 1d39b82:

| Severity | Finding | Fix |
|----------|---------|-----|
| **Critical** | SSH key temp file leak — PEM key persists in /tmp | `applyCredentials` now returns cleanup func; callers defer removal |
| **Critical** | Shell injection in HTTPS credential helper | Replaced shell interpolation with `git-credential-store --file` temp file |
| **Warning** | No HTTP timeout on Vault client | Added `&http.Client{Timeout: 15 * time.Second}` |
| **Warning** | Vault error body leaks internal details | Log body at DEBUG; return only status code to callers |
| **Low** | AppRole JSON built with fmt.Sprintf | Switched to `json.Marshal` |

All tests pass (29 pass, 4 skipped ttyd).

PR #55, opened 2026-03-15
Add stale session cleanup and service recovery

### LINKED ISSUE #21: Agent service: stale session cleanup + service recovery
## Description

Add background cleanup of stale sessions and state reconciliation on service restart. Stopped sessions are auto-destroyed after a configurable TTL, and existing tmux sessions are rediscovered after a crash or restart.

See `services/agents/docs/PLAN.md` (Phase 6) for full details.

## What Gets Built

- Background goroutine for periodic stale session detection
- On startup: reconcile in-memory state with tmux reality (rediscover sessions)
- Config support for `cleanup.stale_session_ttl` and `cleanup.check_interval`
- `test/e2e/cleanup_test.go`

## Acceptance Criteria

- [ ] Stopped sessions auto-destroyed after `stale_session_ttl`
- [ ] After service restart, existing tmux sessions on the agentd socket are rediscovered
- [ ] Rediscovered sessions have correct state (RUNNING vs STOPPED)
- [ ] Cleanup interval is configurable
- [ ] E2e tests verify cleanup and recovery behavior

## Notes

- Blocked by #16 (steel thread)
- Can be parallelized with Phases 4 and 5

dokipen:
## Summary
- Background `Cleaner` goroutine that periodically destroys stopped sessions past a configurable TTL (`cleanup.stale_session_ttl`, default 24h)
- `Manager.RecoverSessions()` on startup rediscovers orphaned tmux sessions after daemon restart, with correct RUNNING/STOPPED state detection
- `CleanupConfig` with `stale_session_ttl` and `check_interval` duration settings
- `StoppedAt` timestamp on sessions for accurate staleness tracking
- E2E tests for cleanup TTL behavior, running session protection, and session recovery
- Fix pre-existing `server.New` call signature in `git_worktree_test.go`

Fixes #21

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] All unit tests pass
- [x] All E2E tests pass (24 pass, 4 skip for ttyd)
- [x] New cleanup tests: stale session destroyed, running session preserved
- [x] New recovery tests: rediscover running sessions, rediscover stopped sessions

### DIFF
--- a/services/agents/internal/config/config.go
+++ b/services/agents/internal/config/config.go
@@ -198,6 +198,10 @@ func validate(cfg *Config) error {
 	}
 
 	// Auth mode validation.
+	if cfg.Cleanup.CheckInterval <= 0 {
+		return fmt.Errorf("cleanup.check_interval must be positive")
+	}
+
 	switch cfg.Auth.Mode {
 	case "none":
 		// ok
--- a/services/agents/internal/config/config_test.go
+++ b/services/agents/internal/config/config_test.go
@@ -3,10 +3,18 @@ package config
 import (
 	"os"
 	"testing"
+	"time"
 )
 
 func validProfiles() map[string]Profile {
 	return map[string]Profile{"test": {Command: "echo test"}}
+}
+
+func validCleanup() CleanupConfig {
+	return CleanupConfig{
+		StaleSessionTTL: 24 * time.Hour,
+		CheckInterval:   5 * time.Minute,
+	}
 }
 
 func TestValidate_ValidLocalhostNoAuth(t *testing.T) {
@@ -14,6 +22,7 @@ func TestValidate_ValidLocalhostNoAuth(t *testing.T) {
 		Host:     "127.0.0.1",
 		Auth:     AuthConfig{Mode: "none"},
 		Profiles: validProfiles(),
+		Cleanup:  validCleanup(),
 	}
 	if err := validate(cfg); err != nil {
 		t.Errorf("expected no error, got: %v", err)
@@ -25,6 +34,7 @@ func TestValidate_ValidLocalhostIPv6(t *testing.T) {
 		Host:     "::1",
 		Auth:     AuthConfig{Mode: "none"},
 		Profiles: validProfiles(),
+		Cleanup:  validCleanup(),
 	}
 	if err := validate(cfg); err != nil {
 		t.Errorf("expected no error, got: %v", err)
@@ -36,6 +46,7 @@ func TestValidate_ValidTokenAuth(t *testing.T) {
 		Host:     "127.0.0.1",
 		Auth:     AuthConfig{Mode: "token", Token: "mysecret"},
 		Profiles: validProfiles(),
+		Cleanup:  validCleanup(),
 	}
 	if err := validate(cfg); err != nil {
 		t.Errorf("expected no error, got: %v", err)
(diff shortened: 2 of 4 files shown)


dokipen (commented):
## Code Review

Build and vet pass cleanly. The overall structure is solid: the channel-based stop/done pattern in `Cleaner` is idiomatic Go, `reconcile` correctly updates both the store and the local pointer so `cleanup()` sees fresh state, and the two-phase config approach (raw strings + parsed durations) cleanly works around YAML's lack of a native duration type.

---

### Critical

**`time.NewTicker` panics on zero or negative interval — no validation in config or `NewCleaner`**

`time.NewTicker(d)` panics with "non-positive interval for NewTicker" when `d <= 0`. There is no guard in `parseDurations`, `validate`, or `NewCleaner`. A user who sets `check_interval: "0s"` or a negative value in YAML will cause the daemon to panic on startup. The E2E tests pass `0` as TTL (intentionally, for immediate expiry), but they also pass `100*time.Millisecond` as interval. The production code path through `config.Load` is the exposure point.

Fix: add to `validate()` in `config.go`:
```go
if cfg.Cleanup.CheckInterval <= 0 {
    return fmt.Errorf("cleanup.check_interval must be positive")
}
if cfg.Cleanup.StaleSessionTTL < 0 {
    return fmt.Errorf("cleanup.stale_session_ttl must be non-negative")
}
```

---

### Warnings

**`Cleaner.Stop()` panics if called twice**

`close(c.stopCh)` panics on a closed channel. In the current `main.go` there is only one call path, but test code using `defer cleaner.Stop()` plus an explicit `cleaner.Stop()` would panic. Consider protecting with a `sync.Once`:
```go
func (c *Cleaner) Stop() {
    c.stopOnce.Do(func() { close(c.stopCh) })
    <-c.doneCh
}
```

**Initial cleanup is delayed by one full `check_interval`**

The `run()` loop starts with a `time.NewTicker` and only fires cleanup after the first tick. At the default 5-minute interval, sessions that are already stale at startup (e.g., pre-existing stopped sessions from before a restart) will sit for 5 minutes before being destroyed. Consider running `c.cleanup()` once immediately before entering the ticker loop, or documenting this intentional delay.

**Recovered sessions have no `AgentProfile` field**

`RecoverSessions` cannot know the original profile, so recovered sessions have `AgentProfile: ""`. This is unavoidable, but it means:
- `manager.List(profileFilter)` with a non-empty filter will silently exclude all recovered sessions.
- Callers relying on `AgentProfile` for routing or display will see an empty string.

A doc comment on `RecoverSessions` noting this limitation would prevent confusion, e.g.: "Note: recovered sessions have AgentProfile set to empty string since the original profile cannot be determined from tmux alone."

**`testSocket` constant duplicates the test config socket name**

`cleanup_test.go:15` declares `const testSocket = "agentd-test"` which hardcodes the same socket as `testdata/config.yaml`. The `TestMain` already loads the config; if the socket name ever changes in the config, the cleanup tests will silently use a different socket and either miss sessions or fail confusingly. Consider deriving the socket from the loaded config, or at minimum adding a comment noting the value must stay in sync.

---

### Suggestions

**`session_lifecycle_test.go` still uses the `"agentd-test"` string literal**

Lines 46 and 365 use the hardcoded string directly rather than the new `testSocket` constant. Now that the constant exists, use it for consistency.

**`RecoverSessions` TOCTOU between `store.List()` and `store.Add()`**

The function builds a `tracked` map from `store.List()` and then calls `store.Add()` in a loop. If `Create` runs concurrently and adds a session between those two calls with the same `TmuxSession` name, `RecoverSessions` will add a duplicate entry. In practice this is safe because `RecoverSessions` is only called at startup before the gRPC server is accepting connections, but a comment noting this assumption would prevent someone from calling it later in the process lifecycle.

---

### Summary

REVIEW STATUS: CHANGES NEEDED
- [Critical] Add positive-interval validation in `config.validate()` to prevent `time.NewTicker` panic on zero/negative `check_interval`
- [Warning] Guard `Cleaner.Stop()` against double-close with `sync.Once`
- [Warning] Document that recovered sessions have empty `AgentProfile`
- [Warning] Remove or sync the `testSocket` constant with the config-loaded value

PR #79, opened 2026-03-15
Disable GraphQL introspection in production

### LINKED ISSUE #48: Auth hardening: disable GraphQL introspection in production
## Description

Discovered during security review of #9. GraphQL introspection is currently enabled unconditionally, including in production. This exposes the full schema to unauthenticated users.

## Acceptance Criteria

- [ ] Disable introspection in production (e.g., via `NODE_ENV` check or explicit config)
- [ ] Keep introspection enabled in development/test environments

dokipen:
## Summary
- Disable Apollo Server introspection when `NODE_ENV=production`
- Remove `__schema` and `__type` from the auth guard's public fields in production, so unauthenticated introspection queries are rejected at both layers

Fixes #48

## Test plan
- [ ] Verify `npm run typecheck` passes
- [ ] Verify `npm run test` passes (31 tests)
- [ ] Manual: start server with `NODE_ENV=production` and confirm introspection query returns an error
- [ ] Manual: start server without `NODE_ENV=production` and confirm introspection still works

### DIFF
--- a/services/issues/src/auth/guard.test.ts
+++ b/services/issues/src/auth/guard.test.ts
@@ -1,0 +1,102 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { parse } from "graphql";
+
+/**
+ * Helper: imports authGuardPlugin with a fresh module graph so the
+ * module-level PUBLIC_FIELDS set picks up the mocked isProduction value.
+ */
+async function loadGuard(isProduction: boolean) {
+  vi.doMock("../env.js", () => ({ isProduction }));
+  const { authGuardPlugin } = await import("./guard.js");
+  return authGuardPlugin;
+}
+
+function makeRequestContext(
+  query: string,
+  currentUser: { id: string } | null = null,
+) {
+  return {
+    document: parse(query),
+    contextValue: { currentUser },
+  } as any;
+}
+
+describe("authGuardPlugin", () => {
+  beforeEach(() => {
+    vi.resetModules();
+  });
+
+  describe("development mode (isProduction=false)", () => {
+    it("allows unauthenticated introspection queries", async () => {
+      const authGuardPlugin = await loadGuard(false);
+      const plugin = authGuardPlugin();
+      const hooks = await (plugin as any).requestDidStart!({} as any);
+      await expect(
+        hooks.didResolveOperation(
+          makeRequestContext("{ __schema { types { name } } }"),
+        ),
+      ).resolves.toBeUndefined();
+    });
+
+    it("allows unauthenticated __type queries", async () => {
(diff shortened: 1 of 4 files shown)


dokipen (commented):
## Code Review

Typecheck: clean. All 31 tests pass.

### Summary

The change correctly implements the acceptance criteria from #48. Introspection is disabled in production at two independent layers: the Apollo Server config and the auth guard. The defense-in-depth approach is sound — if one layer were misconfigured, the other would still block unauthenticated schema exposure.

---

### Critical

None.

---

### Warnings

**`NODE_ENV` check is duplicated across two files**

`services/issues/src/index.ts` line 12 defines `const isProduction = process.env.NODE_ENV === "production"` and uses it, while `services/issues/src/auth/guard.ts` line 10 repeats the raw `process.env.NODE_ENV === "production"` check inline. These two checks must stay in sync forever. If someone later changes the production sentinel value (e.g. to support a `"staging"` environment that also needs introspection disabled), they have to find and update both sites.

A simple fix is to export `isProduction` from a shared config module and import it in both places. Alternatively, `authGuardPlugin()` could accept an `introspection` boolean parameter so the caller (`index.ts`) passes the already-computed flag — this also makes the guard unit-testable with both modes.

**No test coverage for the new behavior**

The existing `auth.test.ts` covers JWT only. There are no tests asserting that the guard allows or blocks `__schema`/`__type` fields depending on `NODE_ENV`. This means the two-layer alignment can silently break. Given the security sensitivity of this change, at least one test exercising the guard in production mode (with `NODE_ENV=production` set) would be valuable.

---

### Suggestions

**`__type` coverage in guard is incomplete**

GraphQL introspection also uses `__typename`, which is valid on any selection set and is handled by the execution engine before field resolvers run. It does not go through root-field dispatch, so its absence from `PUBLIC_FIELDS` is not a real gap here. However, a comment clarifying why `__typename` is not listed (it isn't a root field) would prevent future confusion.

**Apollo Server's `introspection: false` returns a structured error**

When Apollo rejects an introspection query via `introspection: false`, it returns a GraphQL-formatted error before the auth guard's `didResolveOperation` lifecycle hook fires. This means in production the two layers are actually sequential, not truly parallel — Apollo short-circuits first. That's fine for correctness, but worth noting in a comment so the intent of the guard's role is clear.

---

### Alignment assessment

The two layers are logically consistent: both key off `NODE_ENV === "production"` and both target the same introspection surface. There is no gap where one layer would permit what the other blocks. The main risk is the duplicated conditional described in Warnings above.

---

REVIEW STATUS: CHANGES NEEDED
- Extract the shared `isProduction` / `NODE_ENV` check to avoid silent drift between the two files
- Add at least one test covering the guard's behavior in production mode

PR #90, opened 2026-03-15
CLI auth: read PAT from stdin and prompt interactively

### LINKED ISSUE #51: CLI auth: avoid PAT in shell history and process list
## Description

Discovered during security review of #9. The `auth login --pat <token>` flag exposes the PAT in shell history and `ps` output. Consider reading from stdin or a file instead.

## Acceptance Criteria

- [ ] Support reading PAT from stdin (e.g., `echo $TOKEN | issues auth login --pat -`)
- [ ] Consider prompting interactively if no flag provided

dokipen:
## Summary
- Support `--pat -` to read the PAT from stdin, avoiding exposure in shell history and `ps` output
- Prompt interactively for the PAT when no auth flags are provided and stdin is a TTY
- Add `runCliWithStdin` test helper and e2e tests for stdin piping

Fixes #51

## Test plan
- [x] Typecheck passes (`tsc --noEmit`)
- [x] All 11 e2e auth tests pass (including 2 new stdin tests)
- [x] `--pat -` reads token from piped stdin
- [x] Empty stdin with `--pat -` produces clear error message
- [ ] Manual: `echo $TOKEN | issues auth login --pat -` authenticates successfully
- [ ] Manual: `issues auth login` (no flags, TTY) prompts for PAT interactively

### DIFF
--- a/services/issues-cli/src/commands/auth.ts
+++ b/services/issues-cli/src/commands/auth.ts
@@ -23,9 +23,15 @@ const AUTHENTICATE_WITH_PAT = gql`
   }
 `;
 
+const GENERATE_OAUTH_STATE = gql`
+  mutation GenerateOAuthState {
+    generateOAuthState
+  }
+`;
+
 const AUTHENTICATE_WITH_CODE = gql`
-  mutation AuthenticateWithGitHubCode($code: String!) {
-    authenticateWithGitHubCode(code: $code) {
+  mutation AuthenticateWithGitHubCode($code: String!, $state: String!) {
+    authenticateWithGitHubCode(code: $code, state: $state) {
       token
       user {
         id
@@ -130,9 +136,13 @@ export function registerAuthCommand(program: Command): void {
           spinner.succeed(`Authenticated as ${chalk.bold(result.user.login)} (${result.user.displayName})`);
         } else if (options.code) {
           const spinner = ora("Authenticating...").start();
+          const stateData = await client.request<{ generateOAuthState: string }>(
+            GENERATE_OAUTH_STATE
+          );
+          const state = stateData.generateOAuthState;
           const data = await client.request<{ authenticateWithGitHubCode: AuthPayload }>(
             AUTHENTICATE_WITH_CODE,
-            { code: options.code }
+            { code: options.code, state }
           );
           result = data.authenticateWithGitHubCode;
           setAuthToken(result.token);
--- a/services/issues/src/auth/guard.ts
+++ b/services/issues/src/auth/guard.ts
@@ -6,6 +6,7 @@ import type { AuthenticatedContext } from "../schema/resolvers/auth.js";
 // Root fields that don't require authentication.
 // Introspection fields are only public outside production.
 const PUBLIC_FIELDS = new Set([
+  "generateOAuthState",
   "authenticateWithGitHubCode",
   "authenticateWithGitHubPAT",
   ...(isProduction ? [] : ["__schema", "__type"]),
--- a/services/issues/src/auth/state-store.test.ts
+++ b/services/issues/src/auth/state-store.test.ts
@@ -1,0 +1,80 @@
+import { describe, it, expect, vi, afterEach } from "vitest";
+import { OAuthStateStore } from "./state-store.js";
+
+describe("OAuthStateStore", () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it("should generate a hex state token", () => {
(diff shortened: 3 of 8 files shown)


dokipen (commented):
## Security Review: CLI auth stdin PAT (#90)

This PR addresses a real and important security issue — PATs passed directly as `--pat <token>` are visible in shell history and `ps` output. The fix is on the right track. Below are my findings.

---

### Risk Assessment

| Finding | Severity | Location | Recommendation |
|---------|----------|----------|----------------|
| Interactive prompt does not suppress echo | Warning | `auth.ts:86-96` (`promptSecret`) | Use a proper hidden-input mechanism; current `readline` `createInterface` does not disable terminal echo |
| PAT still accepted as a direct CLI argument (`--pat <token>`) | Warning | `auth.ts:113-114` | The original vulnerable path remains available; consider deprecating or removing it |
| `--code <code>` has the same shell-history exposure | Suggestion | `auth.ts:128-136` | Apply the same stdin / prompt pattern to `--code` for consistency |
| Config directory created without explicit mode | Info | `config.ts:9-11` (`ensureConfigDir`) | `mkdirSync` with `recursive: true` uses umask-derived permissions; set `{ mode: 0o700 }` explicitly |
| SQL values interpolated directly in test helper | Info | `helpers.ts:56` | Test-only, but worth using parameterized queries or a proper ORM call to avoid forming a habit |

---

### Detailed Findings

**Warning — `promptSecret` does not hide terminal input**

`promptSecret` at `auth.ts:86-96` uses Node's `readline.createInterface` with `input: process.stdin`. This does **not** disable terminal echo; every character the user types is visible on screen. A PAT entered this way is not protected from shoulder-surfing and may appear in recorded terminal sessions.

The standard approach for hidden input in Node is to set `terminal: false` and drive the TTY directly, or use a library that calls `rl.stdoutMuted` / wraps the stream. A minimal fix that actually suppresses echo:

```typescript
function promptSecret(prompt: string): Promise<string> {
  return new Promise((resolve, reject) => {
    process.stderr.write(prompt);
    const rl = createInterface({ input: process.stdin, output: process.stderr, terminal: true });
    // Mute output so keystrokes are not echoed
    (rl as unknown as { output: { write: (s: string) => void } }).output.write = () => {};
    rl.once("line", (line) => {
      rl.close();
      process.stderr.write("\n"); // move to next line after hidden input
      resolve(line.trim());
    });
    rl.once("error", reject);
  });
}
```

A battle-tested alternative is the `readline` `setRawMode` approach or a small dependency like `read` (npm). This is a **Warning** because users relying on the interactive path receive no actual protection.

**Warning — Direct `--pat <token>` argument still accepted**

`auth.ts:113-114` falls through to use `options.pat` verbatim when it is not `"-"`. The original vulnerability (PAT in shell history and process list) is unchanged for users who do not know about `--pat -`. Consider one of:
- Printing a deprecation warning when a literal token is detected (heuristic: non-`"-"` value that looks like a PAT).
- Removing the literal-value path entirely and requiring `--pat -` or the interactive prompt.

**Suggestion — `--code` has the same exposure**

The OAuth code passed via `--code <code>` is also visible in shell history and `ps`. It has a shorter lifetime than a PAT, but it would be consistent to treat it the same way.

**Info — Config directory permissions**

`config.ts:10`: `mkdirSync(CONFIG_DIR, { recursive: true })` relies on the process umask to restrict access. On systems with a permissive umask (e.g. `0o022`), `~/.issues-cli` will be created as `0o755`, world-readable. Since `auth.json` is written with `0o600`, the file itself is protected, but the directory metadata (existence) is visible. Best practice is `{ recursive: true, mode: 0o700 }`. Low-effort fix.

---

### What the PR Gets Right

- `auth.json` is written with `mode: 0o600` — correct.
- `promptSecret` writes the prompt to **stderr**, not stdout, preventing the prompt from appearing in piped output.
- `readStdin` trims input, preventing accidental newline inclusion in the token.
- Empty-stdin path produces a clear error and non-zero exit code.
- Test coverage for both the happy path and the empty-stdin case is present.
- `runCliWithStdin` in the test helper correctly closes stdin after writing, avoiding hangs.

---

### Summary

The core mechanism (`--pat -` reading from stdin) is implemented correctly and is a meaningful improvement. The two **Warning**-level issues should be resolved before merge: the interactive prompt is not actually hiding input, and the insecure direct-argument path remains available without any warning to users.

The **Info**-level config directory permission finding is a cheap one-line fix that is worth doing in this PR since it touches adjacent code.

Discovered in #90 review.

PR #91, opened 2026-03-15
Optimize Docker image and add npm caching to CI

### LINKED ISSUE #63: Optimize Docker image: prune devDependencies and add npm caching to CI
## Description

Discovered in #62 review. Two optimizations for the issues service Docker/CI setup:

1. **Docker runtime image carries devDependencies** — The Dockerfile copies all `node_modules` from the builder stage. Add `npm ci --omit=dev` in the runtime stage or use a separate production install to reduce image size.

2. **No npm caching in CI** — Add `cache: 'npm'` to `actions/setup-node@v4` to speed up CI runs.

## Acceptance Criteria

- [ ] Runtime Docker image only contains production dependencies
- [ ] CI workflow uses npm cache via `actions/setup-node` cache parameter

dokipen:
## Summary
- Prune devDependencies from runtime Docker image by running `npm ci --omit=dev` in the runtime stage instead of copying all `node_modules` from builder
- Add `cache: 'npm'` to `actions/setup-node@v4` in both CI jobs to speed up CI runs

Fixes #63

## Test plan
- [x] Docker build succeeds locally with the new Dockerfile
- [x] CI workflow syntax validated
- [ ] CI checks pass on PR

### DIFF
--- a/services/issues/Dockerfile
+++ b/services/issues/Dockerfile
@@ -29,4 +29,4 @@ USER issues
 
 EXPOSE 4000
 
-CMD sh -c "npx prisma migrate deploy && npx prisma db seed && node dist/index.js"
+CMD sh -c "npx prisma migrate deploy && node dist/prisma/seed.js && node dist/index.js"
--- a/services/issues/package.json
+++ b/services/issues/package.json
@@ -22,12 +22,12 @@
     "dataloader": "^2.2.3",
     "graphql": "^16.10.0",
     "graphql-scalars": "^1.24.0",
-    "jsonwebtoken": "^9.0.3"
+    "jsonwebtoken": "^9.0.3",
+    "prisma": "^6.5.0"
   },
   "devDependencies": {
     "@types/jsonwebtoken": "^9.0.10",
     "@types/node": "^20.17.0",
-    "prisma": "^6.5.0",
     "tsx": "^4.19.0",
     "typescript": "^5.7.0",
     "vitest": "^3.0.0"


dokipen (commented):
## Performance Review

Overall this is a solid improvement — pruning devDependencies from the runtime image and adding npm caching to CI are both high-value wins. A few issues worth addressing before merge, plus some suggestions for follow-on work.

---

### Findings

| # | Issue | Location | Classification | Recommendation |
|---|-------|----------|----------------|----------------|
| 1 | `npx prisma generate` runs twice, once in builder and once in runtime stage | `Dockerfile:25` | Warning | Copy the generated client from the builder instead of regenerating |
| 2 | `npm ci` in the runtime stage hits the npm registry with no cache mount | `Dockerfile:25` | Warning | Add a BuildKit cache mount: `RUN --mount=type=cache,target=/root/.npm npm ci --omit=dev` |
| 3 | `test-cli` job re-runs `npm ci` and rebuilds the issues service from scratch | `issues-service-ci.yml:49` | Suggestion | Upload the built `dist/` as an artifact in `test-service`, download it in `test-cli` |
| 4 | `docker` job has no BuildKit enabled; cache mounts on `RUN` steps won't activate | `issues-service-ci.yml:53-59` | Suggestion | Add `DOCKER_BUILDKIT=1` env var or `docker buildx build` to the docker job |

---

### Finding 1 — Duplicate `prisma generate` (Warning)

`prisma generate` is run in the builder stage (line 9) and again in the runtime stage (line 25). The builder stage generates `@prisma/client` under `node_modules/@prisma/client`. After `npm ci --omit=dev` installs the production `@prisma/client` package, `prisma generate` must be re-run to populate the query engine binary for the target platform.

This is actually *correct behavior* for a multi-stage build when the builder and runtime stages share the same base (`node:20-alpine`), but it is worth being explicit about why it is necessary. If the two stages ever diverge (e.g. different Alpine versions), the generated binary from the builder would be wrong for the runtime anyway. No change needed, but a comment in the Dockerfile would prevent a future "optimization" that breaks runtime:

```dockerfile
# prisma generate must run here (not copied from builder) so the query engine
# binary matches the production node_modules installed by npm ci --omit=dev
RUN npm ci --omit=dev && npx prisma generate
```

---

### Finding 2 — No npm cache mount in Docker runtime stage (Warning)

The runtime stage runs `npm ci --omit=dev` with no cache. Every `docker build` — including the CI `docker` job — downloads packages from the registry unconditionally. With BuildKit, a cache mount persists the npm cache across builds on the same runner:

```dockerfile
RUN --mount=type=cache,target=/root/.npm \
    npm ci --omit=dev && npx prisma generate
```

This also applies to the builder stage's `npm ci` on line 8. The same mount target can be reused:

```dockerfile
RUN --mount=type=cache,target=/root/.npm npm ci
```

Cache mounts are not baked into the image layer, so they do not affect the final image size.

---

### Finding 3 — `test-cli` redundantly rebuilds the issues service (Suggestion)

`test-cli` depends on `test-service` (`needs: test-service`) but still runs a full `npm ci && npx prisma generate && npm run build` for the issues service (line 49). With the npm cache now in place this is cheaper than before, but the TypeScript compile step is still duplicated. Uploading `dist/` as a GitHub Actions artifact in `test-service` and downloading it in `test-cli` would eliminate the redundant compile.

---

### Finding 4 — BuildKit not explicitly enabled in `docker` job (Suggestion)

The `docker` job (lines 53-59) uses plain `docker build`, which on GitHub-hosted runners defaults to BuildKit since Docker 23, but this is implicit. If Finding 2 is addressed with `RUN --mount=type=cache`, the cache mount only activates under BuildKit. Making it explicit is low-cost insurance:

```yaml
- name: Build Docker image
  env:
    DOCKER_BUILDKIT: 1
  run: docker build services/issues/
```

Or switch to `docker buildx build services/issues/` which always uses BuildKit.

---

### What this PR gets right

- Replacing the copied `node_modules` from the builder with `npm ci --omit=dev` in the runtime stage is the correct approach. It avoids shipping devDependencies (`typescript`, `tsx`, `vitest`, `prisma` CLI, etc.) in the final image. The `prisma` CLI package alone is non-trivial in size.
- Copying `package-lock.json` into the runtime stage (line 23) is required for `npm ci` to work and was correctly added.
- The `cache-dependency-path` values in the CI workflow are correct — `test-service` keys on its own lockfile, `test-cli` keys on both, which produces an appropriate cache per job.

---

**Deferred:**
- Artifact upload/download for `dist/` to eliminate the duplicate build in `test-cli` — new issue recommended, priority:low, discovered in #91 review

dokipen (commented):
Security review for this PR. I looked at both the Dockerfile changes and the CI workflow changes.

---

## Dockerfile Changes

**Critical: `prisma` and `tsx` are devDependencies but required at container runtime**

The CMD runs `npx prisma migrate deploy && npx prisma db seed && node dist/index.js`. The seed script is defined in `package.json` as `"seed": "tsx prisma/seed.ts"`, which means both `prisma` (CLI) and `tsx` are needed at startup. Both are in `devDependencies` and will not be installed by `npm ci --omit=dev`.

Two problems fall out of this:

1. The container will fail to start as-is (the seed step calls `tsx`, which won't be present).
2. If `npx` falls back to fetching the missing packages at container startup time, that is an npm registry call from a production container on every boot — a supply chain risk with no integrity verification and no way to air-gap the deployment.

Either move `prisma` and `tsx` to `dependencies`, or restructure so migrations and seeding run from a separate init container / entrypoint script that is not the production node process. A common pattern is to pre-compile the seed script to JS in the builder stage and run it with `node` in the runtime stage, removing the `tsx` dependency entirely from runtime.

---

**Warning: `npm ci` in the runtime stage opens a second supply chain download window**

The previous approach copied `node_modules` from the builder stage, meaning packages were downloaded and verified exactly once. The new approach runs `npm ci --omit=dev` in the runtime stage, which performs a separate download from the npm registry. If a package is compromised between the two installs, the runtime image gets the compromised version while the builder (where tests ran) got the clean one. The integrity hashes in `package-lock.json` do protect against tampered tarballs, but this is still a wider attack surface than a single install. Consider copying only the production subset of `node_modules` from the builder (e.g. via `npm prune --omit=dev` in the builder before copying) to keep a single verified download.

---

**Warning: Unpinned base image `node:20-alpine`**

Both build stages use `node:20-alpine` without a digest pin. The tag is mutable and can change between builds, meaning two builds from the same commit can produce images with different base OS packages. Pin to a specific digest (e.g. `node:20-alpine@sha256:<digest>`) in both `FROM` lines to make builds reproducible and to prevent a compromised or updated tag from silently introducing vulnerabilities.

---

**Suggestion: `npx prisma generate` runs twice (builder and runtime)**

`npx prisma generate` produces the Prisma client in `node_modules/@prisma/client`. Running it in the builder and then again in the runtime stage is redundant once the generated client is available. If the runtime `npm ci` re-installs `@prisma/client` from scratch, the generated files will not be present anyway until the second `generate` runs, so the double invocation is not harmful — but it is worth noting that the generated client output directory should be copied from the builder or regenerated consistently, not both.

---

## CI Workflow Changes

**Warning: npm cache on a public repository allows cross-branch cache poisoning**

GitHub Actions cache entries are keyed by branch and can be read by any workflow on the same repository. On a public repo a contributor could craft a branch whose `package-lock.json` produces a cache key that collides with a cache entry used by `main`, and then poison that cache with a modified `node_modules`. The `npm ci` command that follows will re-verify package integrity against `package-lock.json` hashes, which mitigates full compromise — but only if every cached file is re-verified. Confirm that `npm ci` with a warm cache still performs integrity checks (it does by default) and document this as an accepted risk, or consider restricting cache writes to trusted branches only via `if: github.ref == 'refs/heads/main'` on a separate cache-warming job.

**Suggestion: `cache-dependency-path` lists both lockfiles for `test-cli` but only one is used per install**

The `test-cli` job lists both `services/issues/package-lock.json` and `services/issues-cli/package-lock.json` as cache dependency paths. This means the cache is invalidated if either lockfile changes, even when only one set of packages actually needs re-downloading. This is the safer choice (no stale cache risk) and is fine to keep, but it is worth being aware that any change to the issues service lockfile will bust the CLI job's cache even if the CLI's own lockfile is unchanged.

---

## Summary

| Finding | Severity | Location |
|---|---|---|
| `prisma` + `tsx` devDependencies required at container startup | Critical | `services/issues/Dockerfile:25,32` |
| Second `npm ci` in runtime stage widens supply chain window | Warning | `services/issues/Dockerfile:25` |
| Unpinned `node:20-alpine` base image | Warning | `services/issues/Dockerfile:1,16` |
| npm cache poisoning possible on public repo | Warning | `.github/workflows/issues-service-ci.yml:28,42` |
| Redundant `prisma generate` in runtime stage | Suggestion | `services/issues/Dockerfile:25` |
| `cache-dependency-path` over-invalidation for `test-cli` | Suggestion | `.github/workflows/issues-service-ci.yml:45-47` |

The Critical finding will cause a runtime breakage and should be resolved before merge. The two Warnings are worth addressing in this PR given the changes are already touching both files.

dokipen (commented):
## Re-review: PR #91 — Optimize Docker image and add npm caching to CI

Re-reviewing after fixes were pushed to address the previously-identified Critical finding.

---

### Critical Finding: RESOLVED

The original critical issue — `prisma` and `tsx` were devDependencies but needed at runtime — is now correctly resolved:

- `prisma` has been moved from `devDependencies` to `dependencies` in `services/issues/package.json`. This means `npm ci --omit=dev` in the runtime image will include it.
- The Dockerfile CMD now calls `node dist/prisma/seed.js` instead of `npx prisma db seed` (which required `tsx` to execute `prisma/seed.ts`). The `tsconfig.json` includes `prisma/**/*` in its compile target with `outDir: dist`, so `dist/prisma/seed.js` will be present after the build stage. This is correct.
- `tsx` remains in `devDependencies` only, which is appropriate since it is no longer needed at runtime.

---

### Current File Assessment

**`services/issues/Dockerfile`**

The approach of copying `package.json` + `package-lock.json*` into the runtime stage and running `npm ci --omit=dev` is sound and correctly sized for production. The `npx prisma generate` in the runtime stage regenerates the Prisma client against the production `node_modules`, which is the right pattern.

One observation worth noting: `npx prisma migrate deploy` runs as the `issues` user inside CMD, but the Prisma migration engine may need write access to the SQLite data directory. The `chown -R issues:issues /app` covers `/app/data`, so this should be fine as long as the database file lives there.

**`services/issues/package.json`**

Clean. `prisma` is now in `dependencies` alongside `@prisma/client`, which is the standard pairing for projects that run migrations at runtime.

**`.github/workflows/issues-service-ci.yml`**

The `cache: 'npm'` additions with explicit `cache-dependency-path` values are correct for a monorepo layout where `npm ci` is run from subdirectory working directories. The `test-cli` job correctly lists both lock files since it installs both services.

---

### Suggestions

**Suggestion — Dockerfile layer ordering:** The `RUN mkdir -p /app/data && addgroup -S ...` step runs after `npm ci --omit=dev`, which is fine. However, if the `npx prisma generate` output ever changes the `node_modules` contents, the `chown` will need to cover those files too. Currently `chown -R issues:issues /app` covers everything under `/app` at that point, so this is fine — just worth keeping in mind if the layer order changes in future.

**Suggestion — `package-lock.json*` glob in COPY:** `COPY --from=builder /app/package-lock.json* ./` uses a glob. If the lock file is absent in the builder stage (e.g. a fresh clone without running `npm install`), this silently copies nothing and `npm ci` will fail with a clear error rather than a silent one. This is a minor robustness note — the builder stage already runs `npm ci` which requires the lock file, so in practice it will always exist. No change needed.

---

### Verdict

The Critical finding from the previous review is fully resolved. The implementation is correct and the approach is clean.

REVIEW STATUS: APPROVED

PR #100, opened 2026-03-15
Auth hardening: session invalidation and token revocation

### LINKED ISSUE #50: Auth hardening: session invalidation and token revocation
## Description

Discovered during security review of #9. Currently there is no mechanism to revoke JWT sessions. A revoked PAT still yields a valid 7-day JWT. The 7-day expiry with no refresh or revocation creates a wide exposure window.

## Acceptance Criteria

- [ ] Implement token revocation (e.g., token blocklist or short-lived tokens with refresh)
- [ ] Consider reducing JWT expiry and adding refresh token flow

dokipen:
## Summary
- Replace long-lived 7-day JWTs with short-lived 15-minute access tokens paired with 30-day refresh tokens
- Add `RefreshToken` and `RevokedToken` database models for server-side session management
- Add `refreshToken` and `logout` GraphQL mutations with refresh token rotation
- Check token blocklist on every authenticated request
- Update CLI with automatic token refresh on 401 and server-side logout

Fixes #50

## Test plan
- [x] All 42 issues service unit tests pass (including new JWT/refresh token tests)
- [x] All 73 CLI e2e tests pass
- [x] TypeScript typecheck passes for both services
- [x] Prisma migration applies cleanly


dokipen (commented):
## Code Review — Auth Hardening: Session Invalidation and Token Revocation

Overall the implementation is solid. The token lifetimes, rotation strategy, and CLI auto-refresh are all well-structured. I found one critical security gap and two warnings that should be addressed before merging.

---

### services/issues/src/schema/resolvers/auth.ts

**Critical: `logout` does not revoke the current access token (jti blocklist gap)**

The `logout` mutation marks the `RefreshToken` row as revoked but never writes to the `RevokedToken` table. Because access tokens are valid for 15 minutes, a user who logs out can continue making authenticated requests until the access token naturally expires. The `RevokedToken` table and the blocklist check in `context.ts` are wired up correctly — they just have no writer.

Fix: at logout time, extract the `jti` from the caller's access token (available from the `Authorization` header on the request context), and insert it into `RevokedToken` with the token's original `expiresAt`. The resolver already receives `prisma`; the context would need to be extended to carry the raw `jti` alongside `currentUser` (or the raw token string), which is a small change.

Alternatively: if a `jti` can't be threaded through cleanly right now, document the gap explicitly and add a tracking issue, but the `RevokedToken` table and blocklist check are dead code until this is wired up, which is misleading.

**Warning: Refresh token rotation is not atomic (TOCTOU race)**

`refreshToken` resolver does a `findUnique`, validates, then does a separate `update` to revoke and a separate `prisma.refreshToken.create`. Under concurrent requests (e.g., two simultaneous retries from the CLI), two threads can both pass the `revoked === false` check before either completes the update, issuing two new token pairs from the same refresh token. Use a transaction:

```ts
const { token, refreshToken: newRefreshToken } = await prisma.$transaction(async (tx) => {
  const updated = await tx.refreshToken.updateMany({
    where: { id: storedToken.id, revoked: false },
    data: { revoked: true },
  });
  if (updated.count === 0) {
    throw new GraphQLError("Invalid or revoked refresh token", {
      extensions: { code: "UNAUTHENTICATED" },
    });
  }
  return issueTokens(tx, storedToken.userId);
});
```

This is particularly important for a security-critical path because token reuse detection depends on atomicity.

**Suggestion: `verifyToken` is imported but never called in `auth.ts`**

Line 8 imports `verifyToken` from `../../auth/jwt.js` but it is not used anywhere in the resolver file. The linter may or may not catch this depending on configuration. Remove the import.

---

### services/issues/src/auth/context.ts

**Suggestion: The `if (jti)` guard is unnecessary**

`JwtPayload.jti` is typed as `string` (non-optional) and `signToken` always sets it via `randomBytes`. The guard `if (jti)` implies it can be falsy, which is inconsistent with the type. If the intent is to tolerate legacy tokens without a `jti` (issued before this PR), that should be documented. Otherwise the check should be removed and `jti` presence should be enforced via `verifyToken` throwing on missing jti.

---

### services/issues/prisma/schema.prisma + migration

**Warning: No expiry-based cleanup for `RevokedToken` or `RefreshToken`**

Both tables will grow unboundedly. `RevokedToken` rows are safe to delete after `expiresAt` (the access token would have expired anyway). `RefreshToken` rows with `revoked: true` or `expiresAt < now` can also be pruned. Without a cleanup mechanism the table will accumulate indefinitely, and the blocklist lookup becomes slower over time (though it is a unique index lookup so it degrades gracefully). This is acceptable for now but should be tracked.

Deferred: new issue recommended — "Add scheduled cleanup for expired/revoked token rows" — priority:low, discovered in #100 review.

---

### services/issues-cli/src/client.ts

**Warning: `getClient()` captures the access token at construction time**

`getClient()` reads `getAuthToken()` once when called. If a command calls `getClient()` and the refresh path in `tryRefresh` writes a new token to disk (via `setAuthTokens`), any subsequent calls to `getClient()` in the same process will still return a client initialized with the old (now-expired) token, causing repeated 401 → refresh cycles. This is low-risk in the current CLI (one command per invocation), but if the CLI is ever extended to issue multiple requests in a pipeline, callers would need to re-call `getClient()` after a refresh. A note in the JSDoc would prevent confusion.

**Suggestion: Silent swallow of refresh errors in `tryRefresh`**

The `catch` block in `tryRefresh` (line 51) discards all errors silently. A debug-level log (gated on `process.env.DEBUG` or similar) would help diagnose cases where refresh fails for reasons other than an expired refresh token (network error, server error, etc.). Not a blocker.

---

### services/issues-cli/src/config.ts

**Suggestion: `setAuthToken` (singular) is now a legacy path that could silently lose a refresh token**

The updated `setAuthToken` spreads `existing` before overwriting `token`, which preserves a stored `refreshToken`. However, the function is still exported and callable. If any future caller uses `setAuthToken` instead of `setAuthTokens` after a fresh login (e.g., a new auth provider), it would silently preserve a stale refresh token. Deprecating or removing `setAuthToken` in favor of `setAuthTokens` would close that path.

---

## Summary

| Finding | Severity |
|---|---|
| `logout` does not write to `RevokedToken` blocklist — access tokens stay valid post-logout | **Critical** |
| Refresh token rotation is not atomic — TOCTOU race under concurrent requests | **Warning** |
| No cleanup for expired/revoked token rows (unbounded growth) | **Warning** (deferred) |
| `getClient()` captures token at construction time | **Warning** |
| Unused `verifyToken` import in `auth.ts` resolver | Suggestion |
| `if (jti)` guard inconsistent with non-optional type | Suggestion |
| Silent error swallow in `tryRefresh` | Suggestion |
| `setAuthToken` (singular) is a footgun for future callers | Suggestion |

REVIEW STATUS: CHANGES NEEDED
- `logout` mutation must write the access token's `jti` to `RevokedToken` (Critical — otherwise the blocklist is wired but never written)
- Refresh token rotation must be atomic using `prisma.$transaction` (Warning — security-critical race condition)
- `getClient()` token capture timing should be documented or fixed (Warning)

dokipen:
## Tester Review — PR #100 Auth Hardening

### Build Status

Both `tsc --noEmit` checks pass clean. All 115 tests pass:
- **issues service**: 42/42 unit tests pass
- **issues-cli**: 73/73 e2e tests pass

---

### Coverage Assessment by Feature

#### `jwt.ts` — `signToken` / `verifyToken` / `generateRefreshToken`

Well covered. Tests verify: sign/verify round-trip, jti uniqueness, 15-minute expiry, rejection of invalid/expired/wrong-secret tokens, refresh token length and entropy. Nothing critical missing here.

#### `guard.ts` — `authGuardPlugin`

Covered for the main cases. One gap noted below.

#### `context.ts` — `buildAuthContext` (revocation blocklist check)

**No unit tests exist for this module.** This is the highest-value new code in the PR — it enforces the blocklist on every authenticated request — and it has zero direct test coverage.

#### `auth.ts` resolvers — `refreshToken` / `logout` mutations

**No unit tests exist for these resolvers.** The e2e suite only covers `logout` with a passing happy-path check ("should succeed even without stored token"). The `refreshToken` mutation has no e2e test at all, and the resolver's error branches are untested.

#### CLI `client.ts` — auto-refresh on 401

**No unit tests exist for the auto-refresh intercept logic.** The e2e suite never exercises the refresh path because it uses long-lived (1-hour) test tokens.

---

### Findings

**Warning — `buildAuthContext` has no tests**

`/services/issues/src/auth/context.ts` is the enforcement point for the entire blocklist feature. The following behaviors are not tested by any test:

- A revoked jti is rejected (token is in `RevokedToken` table → returns `null` user)
- A valid, non-revoked token is accepted through the blocklist check
- A missing/malformed Authorization header returns `null` user
- A token with a valid signature but a deleted user returns `null` user
- The `jti`-missing guard branch (`if (jti)`) — a token without a jti currently bypasses blocklist checks entirely

The last point is particularly worth noting: `verifyToken` always sets `jti` from the payload, but the guard `if (jti)` in `buildAuthContext` would silently skip revocation for a legacy token with no jti field. There is no test verifying this edge case behaves safely.

**Warning — `refreshToken` and `logout` resolvers have no unit tests**

`/services/issues/src/schema/resolvers/auth.ts` contains four new/modified resolvers. The following branches have no test coverage:

- `refreshToken` happy path (returns new token + rotated refresh token)
- `refreshToken` with an unknown token → `"Invalid or revoked refresh token"` error
- `refreshToken` with a revoked token → same error
- `refreshToken` with an expired token → `"Refresh token expired"` error
- `logout` with a valid refresh token (server-side revocation actually runs)
- `logout` with an already-revoked token (no-ops correctly)

The single e2e logout test calls `auth logout` after the test suite has already cleared the refresh token, so server-side revocation is never exercised.

**Warning — CLI auto-refresh path (`client.ts`) is never exercised**

The e2e helper at `/services/issues-cli/test/e2e/helpers.ts` signs a 1-hour JWT:

```ts
const authToken = jwt.sign({ userId: TEST_USER_ID, jti }, TEST_JWT_SECRET, { expiresIn: "1h" });
```

Because the test token never expires during a test run, the `tryRefresh` path in `client.ts` is dead during all 73 e2e tests. Missing coverage:

- A 401/UNAUTHENTICATED response triggers a refresh attempt
- After a successful refresh, the original request is retried with the new token
- When no refresh token is stored, the original error propagates unchanged
- When the refresh itself fails, the original error propagates unchanged

**Suggestion — `authGuardPlugin` missing a mixed-operation test**

The guard comment says it "prevents bypass via anonymous operations or mixed queries", but there is no test for an operation that mixes a public field with a protected field (e.g., `{ authenticateWithGitHubPAT(...) { token } me { id } }`). The implementation loops over all selections and throws on the first non-public field, so the behavior is correct, but the intent is unverified.

**Suggestion — `logout` e2e test doesn't verify server-side session invalidation**

The e2e test only checks the exit code and "Logged out" message. It does not verify that a subsequent authenticated request using the old token is rejected. A test that: (1) logs out, (2) issues a request with the old access token, would confirm the full session-invalidation contract — which is the core security promise of this PR.

**Suggestion — `RevokedToken` table is never populated in tests**

The `RevokedToken` model exists in the schema and is queried in `buildAuthContext`, but the `logout` mutation only updates `RefreshToken.revoked`; it does not write to `RevokedToken`. There are no tests that write to `RevokedToken` directly either. If the intent is to also invalidate the outstanding access token (not just the refresh token) at logout time, a row needs to be inserted into `RevokedToken` with the access token's `jti`. This is a potential logic gap in the implementation that tests would catch.

---

### Summary

| Finding | Classification |
|---|---|
| `buildAuthContext` (blocklist enforcement) has no tests | Warning |
| `refreshToken` / `logout` resolvers have no unit tests | Warning |
| CLI auto-refresh path is never exercised by e2e tests | Warning |
| Missing mixed-operation guard test | Suggestion |
| `logout` e2e doesn't verify post-logout requests are rejected | Suggestion |
| `RevokedToken` table appears unused at logout — access tokens not invalidated | Suggestion |

PR #138, opened 2026-03-15
Add Issues UI plan and user stories

dokipen:
## Summary

- Adds `services/issues-ui/docs/PLAN.md` with full architecture decisions, directory structure, infrastructure changes, E2E testing strategy, GraphQL queries, and phased implementation plan
- Adds 6 user story documents covering authentication, kanban board, project selector, ticket detail, GitHub OAuth, and deployment/CI

## Context

Closes #137 — planning issue for the Issues UI read-only kanban board milestone (21 story points across 6 phases).

Follow-on phase tickets will be created after this PR merges.

## Test plan

- [ ] Review plan document for completeness and accuracy
- [ ] Verify user stories match the plan's phase descriptions
- [ ] Confirm GraphQL queries match the existing schema in `services/issues/src/schema/typeDefs/`

### DIFF
--- a/services/issues-ui/docs/PLAN.md
+++ b/services/issues-ui/docs/PLAN.md
@@ -155,16 +155,23 @@ export default defineConfig({
 
 ### Auth Fixture
 
-JWT payload matches `services/issues/src/auth/jwt.ts` — `{ userId, jti }` signed with `JWT_SECRET`:
+The access token is a JWT matching `services/issues/src/auth/jwt.ts` — `{ userId, jti }` signed with `JWT_SECRET`:
 
 ```ts
 const token = jwt.sign({ userId: "e2e-test-user", jti: "e2e" }, "e2e-test-secret", { expiresIn: "1h" });
 ```
 
+The refresh token is an opaque hex string (not a JWT). The seed script must insert a `RefreshToken` row via Prisma directly — `generateRefreshToken()` produces a random 64-char hex string, so the fixture creates one and stores it in the database alongside the test user.
+
+### Database Setup
+
+`global-setup.ts` must run `prisma migrate deploy` against the test database before seeding. The `webServer` `npm start` command alone does not migrate a fresh `test.db`.
+
 ### Test Seed Data
 
-`services/issues/prisma/seed-e2e.ts` creates:
+`services/issues/prisma/seed-e2e.ts` creates via direct Prisma inserts:
 - 1 user (`e2e-test-user` / `e2e-tester`)
+- 1 refresh token row linked to the test user
 - 1 project (`E2E Test Project`)
 - 4+ tickets (one per state), with labels, assignee, story points
 - Comments on tickets
@@ -284,13 +291,15 @@ Expanded ticket view with all fields, comments, and blocking relationships.
 
 Full GitHub OAuth redirect login alongside PAT login.
 
+> **Note:** Production OAuth requires Caddy's `try_files` to serve `index.html` for the `/auth/callback` route. This Caddyfile change lands in Phase 5. OAuth can be developed and tested locally (Vite handles SPA routing in dev), but won't work in production until Phase 5 deploys.
+
 **Delivers:**
 - `auth/AuthCallback.tsx` — Route at `/auth/callback`
 - Updated `auth/LoginPage.tsx` — "Sign in with GitHub" button
 - `VITE_GITHUB_CLIENT_ID` env var for build-time OAuth client ID
 - `.env.example` documenting the variable
 
-**Verify:** Full OAuth redirect flow works end-to-end. PAT login still works as fallback.
+**Verify:** Full OAuth redirect flow works end-to-end (locally). PAT login still works as fallback.
 
 ### Phase 5: Caddy config + CI + deploy + polish (est: 5)
 **Blocked by:** Phase 3, Phase 4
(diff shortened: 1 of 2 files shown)


dokipen (commented):
## Plan Review: Issues UI Read-Only Kanban Board

Overall this is a solid, well-structured planning document. The architecture choices are appropriate, the phasing is logical, and the plan shows real familiarity with the existing codebase. A few corrections and gaps are noted below.

---

### Critical

**1. Refresh token is not a JWT — the E2E auth fixture approach is wrong**

`services/issues/src/auth/jwt.ts` uses a random 32-byte hex string for refresh tokens (`generateRefreshToken()`), not a JWT. Only the access token is a JWT with `{ userId, jti }`. The plan's E2E auth fixture correctly signs the access token:

```ts
const token = jwt.sign({ userId: "e2e-test-user", jti: "e2e" }, "e2e-test-secret", { expiresIn: "1h" });
```

But the seed script must also insert a corresponding refresh token row into the database (via Prisma), not synthesize one from JWT. The `seed-e2e.ts` must create the `RefreshToken` record directly so that auto-refresh during E2E tests works at all. The plan does not address this.

**2. `ci-status` gate job must be updated when the new CI jobs are added**

The `ci-status` job in `.github/workflows/ci.yml` (lines 116–151) has a hardcoded `needs` list and a loop over specific result variables. Adding `issues-ui-ci` without updating `ci-status` to include it will allow the gate to pass even if the new job fails. The plan's CI section must explicitly call this out as a required change.

---

### Warnings

**3. GraphQL query for board tickets requests `blockedBy { id }` — schema requires a full `Ticket` selection set**

The `BoardTickets` query in the plan selects `blockedBy { id }`. The schema defines `blockedBy` as `[Ticket!]!`, meaning `id` is valid. This is fine at the GraphQL level, but it is inconsistent with the `TicketDetail` query which selects `blockedBy { id, title, state }`. More importantly, since `Ticket.number` is a required field (`Int!`), any fragment reuse that omits `number` will cause issues in typed codegen if the project later adopts it. Not a bug today, but worth flagging.

**4. The `projects` query returns `[Project!]!` — `repository` field is not nullable**

The plan's projects query `query { projects { id, name, repository } }` matches the schema exactly. However, `Project.repository` is `String!` in the schema, so selecting it from the UI without null-guards is safe. No issue here — confirming accuracy.

**5. Missing Caddy route for `/auth/callback`**

The OAuth `AuthCallback` component handles `/auth/callback`. Caddy's current fallback `handle` block responds with a plain-text string. After the Caddyfile update the fallback becomes `file_server` with `try_files`, which will correctly serve `index.html` for `/auth/callback`. This is fine — but the plan should note that the Caddyfile change in Phase 5 is a prerequisite for the OAuth flow in Phase 4 to work in production. As written, Phase 4 is listed as independent from Phase 5, which could mislead an implementer.

**6. `install-caddy-site.sh` must be updated alongside `Caddyfile`**

The plan mentions updating both `infrastructure/Caddyfile` and `infrastructure/install-caddy-site.sh`. This is correct — reading `install-caddy-site.sh` confirms it embeds its own copy of the site block as a heredoc (lines 50–78). The plan's wording "replace the fallback `handle` block in both" is accurate but brief. Phase 5 should explicitly list this as a deliverable so it isn't forgotten.

**7. `@types/jsonwebtoken` missing from dev dependencies**

The plan lists `jsonwebtoken` as a dev dependency for the E2E fixture. Since this is TypeScript, `@types/jsonwebtoken` is also needed. Minor, but the dependency list in the plan is the input for the Phase 0 `package.json`.

**8. The `webServer` config in `playwright.config.ts` does not account for Prisma migrations**

The E2E `webServer` command `cd ../issues && DATABASE_URL=file:./test.db JWT_SECRET=e2e-test-secret npm start` assumes the test database is already migrated. The plan should specify that `global-setup.ts` (or a pre-start hook) runs `prisma migrate deploy` against `test.db` before seeding, otherwise the first run against a clean environment will fail.

---

### Suggestions

**9. Token storage in `localStorage` — worth noting the XSS trade-off**

The plan stores tokens in `localStorage` (`cadence_token`, `cadence_refresh_token`). For a self-hosted internal tool this is a pragmatic choice, but it is worth a one-line note in the plan or `01-authentication.md` acknowledging the trade-off (vs. `httpOnly` cookies), so the decision is clearly intentional rather than an oversight.

**10. `first: 100` for BACKLOG/REFINED/IN_PROGRESS with no pagination UI**

The plan caps these columns at 100 tickets. For a small team this is fine. A brief note that this is a deliberate simplification (and that the `tickets` query supports cursor pagination via `after`/`pageInfo` for a future enhancement) would make the design decision self-documenting.

**11. Auto-refresh `setInterval` without cleanup**

Phase 5 mentions a 60-second `setInterval` for board auto-refresh. The `useTickets` hook should clear the interval on unmount. This is standard React hooks hygiene — worth a one-line note in Phase 5's deliverables or in `02-kanban-board.md`.

---

### Summary

The GraphQL queries, JWT payload shape, token expiry values, and auth mutation names all match the existing schema and `jwt.ts` implementation accurately. The phasing and parallel-execution model (Phases 2–3 and Phase 4 in parallel, converging at Phase 5) is sound. The three Critical/Warning items above should be resolved before implementation begins — particularly the refresh token seeding approach and the `ci-status` gate update.

---

REVIEW STATUS: CHANGES NEEDED
- Refresh token E2E seeding strategy needs to account for the opaque (non-JWT) refresh token requiring a Prisma insert
- `ci-status` gate job update must be called out explicitly in the CI plan
- Phase 4 / Phase 5 ordering dependency on Caddyfile for OAuth callback must be documented

dokipen (commented):
All review findings addressed in 744ab8d:

- **Refresh token seeding** (Critical): Documented that refresh tokens are opaque hex strings requiring direct Prisma inserts, not JWTs. Added RefreshToken row to seed data list.
- **`ci-status` gate** (Critical): Explicitly listed in Phase 5 deliverables.
- **Phase 4/5 dependency** (Warning): Added note that production OAuth requires Phase 5's `try_files` Caddyfile change.
- **Prisma migration** (Warning): Added requirement for `global-setup.ts` to run `prisma migrate deploy`.
- **`@types/jsonwebtoken`** (Warning): Added to dev dependency list.
- **localStorage trade-off** (Suggestion): Added note in auth user story.

Ready for merge.

PR #774, opened 2026-04-10
feat(#665): support CADENCE_* env vars for per-machine ticket provider config

dokipen:
## Summary
- `detect-provider.sh` now reads `CADENCE_PROVIDER`, `CADENCE_PROJECT_ID`, and `CADENCE_API_URL` env vars, each overriding the corresponding `CLAUDE.md` field
- `CADENCE_API_URL` takes precedence over the legacy `ISSUES_API_URL` (both remain supported for backward compatibility)
- `README.md` gains a "Per-Machine Ticket Provider Overrides" subsection documenting the `~/.claude/settings.json` env block pattern
- `skills/ticket-provider/SKILL.md` override note updated to reference `CADENCE_*` vars alongside `ISSUES_API_URL`

Fixes #665

## Test plan
- [x] Shellcheck passes
- [x] Smoke-tested all three env var overrides and precedence (`CADENCE_API_URL` > `ISSUES_API_URL` > `CLAUDE.md`)
- [x] Unset vars fall back to `CLAUDE.md` values unchanged

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,22 @@ go vet ./... && go test ./...
 
 The agents and scripts will use this command automatically.
 
+### Per-Machine Ticket Provider Overrides
+
+To override ticket provider settings on a specific machine without modifying tracked files, set `CADENCE_*` env vars in `~/.claude/settings.json`:
+
+```json
+{
+  "env": {
+    "CADENCE_PROVIDER": "issues-api",
+    "CADENCE_API_URL": "https://cadence.example.com",
+    "CADENCE_PROJECT_ID": "my-project"
+  }
+}
+```
+
+These take precedence over `CLAUDE.md` and apply across all worktrees and clones on the machine. Any combination of the three vars can be set — only the fields you specify are overridden. `CADENCE_API_URL` is also the canonical replacement for the legacy `ISSUES_API_URL` env var (both are supported for backwards compatibility).
+
 ## Adding Project-Specific Agents
 
 Add domain-specific agents to your project's `.claude/agents/` directory. They layer on top of the plugin's core agents:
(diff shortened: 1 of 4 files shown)


dokipen (commented):
## Code Review

| Warning | `commands/lead/SKILL.md:72-75` | Still documents `ISSUES_API_URL` as the override mechanism; `CADENCE_API_URL` not mentioned | Update the QA override example to show `CADENCE_API_URL` (with `ISSUES_API_URL` noted as legacy) |
| Suggestion | `skills/ticket-provider/scripts/detect-provider.sh:43-44` | `CADENCE_PROVIDER` env var can be set to an invalid value (e.g. `"typo"`) with no validation; downstream errors will be confusing | Low priority; add a guard or let it fail at usage time |

---

### Warning: Stale `ISSUES_API_URL` reference in `commands/lead/SKILL.md`

`skills/ticket-provider/SKILL.md` was updated to reference `CADENCE_API_URL` as the canonical var. However `commands/lead/SKILL.md` lines 72–75 still instruct users to use `ISSUES_API_URL` exclusively, without mentioning `CADENCE_API_URL`. Users reading the `/lead` skill docs will miss the new canonical form. The fix is a one-liner: swap the example to `CADENCE_API_URL` and add a note that `ISSUES_API_URL` is a legacy alias.

---

**Shell correctness:** The precedence chain `${CADENCE_API_URL:-${ISSUES_API_URL:-$API_URL}}` is correct. Empty-string env vars correctly fall back to CLAUDE.md values via `:-` expansion — confirmed by local testing.

**Shellcheck:** Passes with no warnings.

**Documentation accuracy:** README and `SKILL.md` additions are accurate. The `~/.claude/settings.json` env block pattern is well-documented and the partial-override behavior is clearly called out.

**Edge cases:** Empty strings handled correctly by `:-`. No issues with unset vars. No injection risk since values are passed via `jq --arg`.

---

REVIEW STATUS: CHANGES NEEDED
- Update `commands/lead/SKILL.md` to reference `CADENCE_API_URL` (with `ISSUES_API_URL` as legacy alias) in the QA override example

dokipen (commented):
Fixed the missed reference in `commands/lead/SKILL.md` — updated the QA/local override example to use `CADENCE_API_URL` (with a note that the legacy `ISSUES_API_URL` is still supported). Pushed in follow-up commit 6a2f75b.

PR #778, opened 2026-04-10
feat(#668): global default fallback in ~/.claude/cadence.json

dokipen:
## Summary

- `detect-provider.sh` now reads top-level `provider`/`project_id`/`api_url` fields from `~/.claude/cadence.json` as a global default when the project's `CLAUDE.md` has no `## Ticket Provider` section
- Full precedence order: per-repo override > `CLAUDE.md` > global default > built-in `github`
- Updated `SKILL.md` and `README.md` to document the complete `~/.claude/cadence.json` format

Closes #668

## Test plan

- [ ] No config anywhere → `provider: github` (default unchanged)
- [ ] Global default in `~/.claude/cadence.json`, no CLAUDE.md provider section → global values used
- [ ] `CLAUDE.md` has provider config + global default exists → `CLAUDE.md` wins
- [ ] Per-repo override + `CLAUDE.md` provider config → per-repo wins


dokipen (commented):
## Code Review — PR #778

| Warning | `detect-provider.sh:69` | Global fallback fires on empty `PROVIDER` regardless of whether `## Ticket Provider` section exists in CLAUDE.md | Track section presence with a sentinel variable; only apply fallback when no section was found |
| Warning | `detect-provider.sh:73-75` | When CLAUDE.md has a partial `## Ticket Provider` section (e.g. `project_id` set but no `provider:`), the global block runs and can overwrite the already-set `PROJECT` with the global default | Restrict the global block to cases where the section is absent, not just where `PROVIDER` is empty |
| Suggestion | `README.md:140` | "...it belongs in `CLAUDE.md` or the global default" — sentence is ambiguous about whether per-repo `api_url` override is allowed (it is not, per the script) | Reword to clearly separate the two distinct cases: per-repo overrides can't carry `api_url`; global defaults can |

---

### Warning — Incorrect trigger condition for global fallback (lines 69, 73-75)

The documented behaviour is: _"Global defaults apply when CLAUDE.md has no `## Ticket Provider` section."_ The implementation instead triggers on `[ -z "$PROVIDER" ]`, which is also true when a `## Ticket Provider` section **exists** but omits the `provider:` field.

Example that breaks the intended precedence:

```markdown
## Ticket Provider
project_id: repo-specific-project
```

With `~/.claude/cadence.json` containing `{"provider":"issues-api","project_id":"global-project","api_url":"https://..."}`:

- `PROVIDER` is empty after awk → global block fires
- `PROJECT` is `"repo-specific-project"` after awk → line 74 overwrites it with `"global-project"` (because `_global_project` is non-empty)
- Final result uses the global `project_id`, discarding what CLAUDE.md explicitly set

The fix is to track whether the `## Ticket Provider` section was found in awk, emit a sentinel (e.g. add a `found` flag as a 4th tab-separated field), and key the global-fallback block on that flag rather than on `PROVIDER` being empty.

```awk
END { print provider "\t" project "\t" api_url "\t" in_section_ever }
```

```sh
_section_found=$(printf '%s' "$_parsed" | cut -f4)
if [ -z "$_section_found" ] && [ -f "$_cadence_cfg" ]; then
  ...
fi
```

(Note: `in_section` is 0 at END because it resets on `^## `. A separate `found` flag set to 1 on match and never cleared is needed.)

---

**Deferred:**
- README `api_url` sentence ambiguity → new low-priority issue, documentation-only change, discovered in PR #778 review

---

**Shellcheck:** passes cleanly. No backtick substitutions, no unquoted variables.

REVIEW STATUS: CHANGES NEEDED
- Global fallback condition should key on section-absent sentinel rather than empty `PROVIDER`, to avoid overwriting CLAUDE.md-set `project_id`/`api_url` when only `provider:` is missing from the section

dokipen (commented):
Test comment to verify posting works

dokipen (commented):
Fixed the sentinel bug identified in review. Now using a 4th awk output field (`found`) to detect whether `## Ticket Provider` section was present at all, rather than checking if `PROVIDER` is empty. This correctly handles partial sections (e.g. `project_id` set but no `provider:`). All 5 scenarios pass.

PR #780, opened 2026-04-19
feat(#670): mobile dialog for terminal text input

dokipen:
## Summary

Adds a "Type…" button to the `MobileSessionView` header. Tapping it opens a native `<dialog>` modal with a textarea, allowing users to compose text comfortably above the on-screen keyboard. Submitting sends the text to the terminal (appended with `\r`). Shift+Enter inserts a newline; Enter alone submits. Cancel and backdrop-click dismiss without sending.

Desktop/non-mobile viewports are unaffected — `MobileSessionView` is rendered only on mobile.

Closes #670

## Test plan

- [x] `npm run test` in `services/issues-ui` — all 783 tests pass
- [x] New tests cover: open dialog, submit text, dismiss after submit, cancel without sending, Enter key submit, Shift+Enter no-op, backdrop-click dismiss

N/A — UI-only change. No new API endpoints, auth boundaries, or external inputs. Text entered in the dialog is forwarded to the existing `sendInput` terminal handle, which has no change in trust boundary.

### DIFF
--- a/services/issues-ui/src/components/MobileSessionView.test.tsx
+++ b/services/issues-ui/src/components/MobileSessionView.test.tsx
@@ -1,6 +1,6 @@
 // @vitest-environment jsdom
 import { describe, it, expect, vi, afterEach, beforeEach } from "vitest";
-import { render, fireEvent, act, cleanup } from "@testing-library/react";
+import { render, screen, fireEvent, act, cleanup } from "@testing-library/react";
 import { forwardRef, useImperativeHandle } from "react";
 import { create } from "@bufbuild/protobuf";
 import { SessionSchema } from "../gen/hub/v1/hub_pb";
@@ -27,9 +27,16 @@ vi.mock("../styles/agents.module.css", () => ({
     mobileHeader: "mobileHeader",
     mobileSessionContent: "mobileSessionContent",
     mobileBackButton: "mobileBackButton",
+    mobileTypeButton: "mobileTypeButton",
     mobileEnterButton: "mobileEnterButton",
     mobileEscButton: "mobileEscButton",
     mobileCloseButton: "mobileCloseButton",
+    mobileInputDialog: "mobileInputDialog",
+    mobileInputDialogContent: "mobileInputDialogContent",
+    mobileInputTextarea: "mobileInputTextarea",
+    mobileInputDialogActions: "mobileInputDialogActions",
+    mobileInputCancelButton: "mobileInputCancelButton",
+    mobileInputSubmitButton: "mobileInputSubmitButton",
   },
 }));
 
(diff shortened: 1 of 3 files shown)


dokipen (commented):
## Code Review — PR #780: mobile dialog for terminal text input

All 776 tests pass (confirmed locally via `npm run test` in `services/issues-ui`). Test bar met per `services/issues-ui/CLAUDE.md`.

---

### Summary

| Warning | `MobileSessionView.tsx` line 33–35 | Effect cleanup fires on every `open` change, not only on unmount — close logic is split between cleanup and effect body, creating a fragile ordering dependency | Move the `el.close()` call exclusively into the effect body; use cleanup only for unmount safety |
| Suggestion | `MobileSessionView.tsx` line 190–195 | `onSubmit` inline arrow creates a new function each render | Wrap in `useCallback` for consistency with the rest of the file, or accept it as-is given `MobileInputDialog` isn't memoized |
| Suggestion | `MobileSessionView.test.tsx` line 282–297 | Backdrop-click test fires `fireEvent.click(dialog)` on the dialog element — passes for the right mechanical reason but doesn't distinguish a true backdrop click from a click on the dialog border itself | Document the limitation in a comment, or accept as adequate for jsdom |
| Suggestion | `MobileSessionView.tsx` line 22 | `text` state is not cleared when `open` is forced to `false` externally (bypassing `handleCancel`/`handleSubmit`) — stale draft text would reappear on next open | Add a `useEffect([open])` to `setText("")` whenever `open` transitions to `false` |

---

### Warning Detail

**Effect cleanup / close logic split** (`MobileSessionView.tsx` lines 24–36)

```ts
useEffect(() => {
  const el = dialogRef.current;
  if (!el) return;
  if (open && !el.open) {
    el.showModal();
    textareaRef.current?.focus();
  } else if (!open && el.open) {
    el.close();           // ← path A: intentional close
  }
  return () => {
    if (el.open) el.close();  // ← path B: also fires on every open→false transition
  };
}, [open]);
```

When `open` flips `true → false`, React runs the **previous** effect's cleanup (path B) before running the new effect body (path A). Path B fires `el.close()` first (guard passes because `el.open` is still `true`). By the time path A runs, `el.open` is already `false`, so the explicit `el.close()` in the effect body becomes a silent no-op. The current `el.open` guard prevents a crash, but the structure relies on React's cleanup-before-effect ordering to avoid a double-close, and the intentional `el.close()` in the effect body is never the one that actually executes.

Idiomatic fix — put all close logic in the effect body and use cleanup only for unmount:

```ts
useEffect(() => {
  const el = dialogRef.current;
  if (!el) return;
  if (open) {
    if (!el.open) {
      el.showModal();
      textareaRef.current?.focus();
    }
  } else {
    if (el.open) el.close();
  }
  return () => {
    // unmount safety only
    if (el.open) el.close();
  };
}, [open]);
```

This is semantically equivalent today, but explicit about intent and robust against future refactors that might change the `open` dependency.

---

### Suggestions (one-liners)

- **Inline `onSubmit` handler**: Wrap in `useCallback` at the `MobileSessionView` level for stylistic consistency — `onSubmit` and `onClose` are created fresh each render while the dialog itself uses `useCallback` throughout.
- **Backdrop-click test**: Add a comment noting that `fireEvent.click(dialog)` simulates clicking the dialog element directly, which is how jsdom expresses a backdrop click (no child targeted), so the test is valid but limited to that mechanic.
- **Stale `text` on forced close**: Add `useEffect(() => { if (!open) setText(""); }, [open])` to clear draft text if `open` is ever set to `false` externally, preventing stale text from reappearing on the next open.

---

REVIEW STATUS: CHANGES NEEDED
- Effect cleanup / close logic split (Warning above) — low-risk but worth fixing for code clarity before merge

dokipen (commented):
Addressed code review feedback:

- **Stable callbacks**: extracted `onSubmit`/`onClose` inline arrows into `useCallback` handlers (`handleDialogSubmit`, `handleDialogClose`) in `MobileSessionView`, consistent with the `useCallback`-heavy style throughout the file
- **Stale text guard**: added `useEffect(() => { if (!open) setText(""); }, [open])` in `MobileInputDialog` — clears draft text whenever the dialog closes, including external/unmount paths

The effect cleanup / close logic split (warning) is intentional and mirrors the pattern in `ConfirmDialog.tsx` — the cleanup function serves as an unmount safety guard; the `else if (!open && el.open)` branch handles the `open: true → false` transition in the normal render path.

PR #783, opened 2026-04-20
fix(#673): use carriage return for text input Enter in notification dropdown

dokipen:
## Summary

- Changed notification dropdown text input to append `\r` instead of `\n` when submitting text to an agent session
- `\r` (carriage return) is the correct terminal Enter key — consistent with how the mobile dialog sends terminal input via the WebSocket path
- `\n` may not register as Enter in raw-mode terminal prompts (e.g., Claude Code's interactive prompts running in cbreak/raw mode)
- Updated tests to assert `\r` is sent

Fixes #673

## Test plan

- [x] `npm run test` — 786 tests pass
- [x] `shellcheck commands/**/*.sh skills/**/*.sh` passes
- [ ] Manual testing: send text via the notification dropdown to a waiting agent session

N/A — changes only the newline character appended to user-submitted text; no new trust boundaries or external inputs introduced.

### DIFF
--- a/services/issues-ui/src/components/NotificationDropdown.test.tsx
+++ b/services/issues-ui/src/components/NotificationDropdown.test.tsx
@@ -418,7 +418,7 @@ describe("NotificationDropdown — text input prompt", () => {
     expect(getByTestId("btn-send")).toBeTruthy();
   });
 
-  it("Send button calls sendSessionInput with typed text + \\n", async () => {
+  it("Send button calls sendSessionInput with typed text + \\r", async () => {
     const agentName = "lead";
     const sessionId = "sess-text";
     const sessions = [
@@ -437,7 +437,7 @@ describe("NotificationDropdown — text input prompt", () => {
     fireEvent.click(getByTestId("btn-send"));
 
     await vi.waitFor(() => {
-      expect(sendSessionInput).toHaveBeenCalledWith(agentName, sessionId, "typed text\n");
+      expect(sendSessionInput).toHaveBeenCalledWith(agentName, sessionId, "typed text\r");
     });
   });
 });
@@ -719,7 +719,7 @@ describe("NotificationDropdown — select prompt delta variants", () => {
 
 
 describe("NotificationDropdown — text input Enter key submission", () => {
-  it("pressing Enter in the text input calls sendSessionInput with typed text + \\n", async () => {
+  it("pressing Enter in the text input calls sendSessionInput with typed text + \\r", async () => {
     const agentName = "lead";
     const sessionId = "sess-enter";
     const sessions = [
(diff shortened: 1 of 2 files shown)


dokipen (commented):
## Code Review

| Warning | `NotificationDropdown.tsx:137,145` | yesno buttons send `"y\n"` / `"n\n"` via the same REST → PTY write path; `\n` does not register as Enter in raw/cbreak mode | Change to `"y\r"` / `"n\r"` — same fix applied here to text input |

### Warning: yesno buttons use `\n` on a raw PTY path

The REST → `PTYManager.WriteInput` → `sess.master.Write` path writes bytes verbatim to the PTY master — no line-discipline INLCR/ICRNL translation applies to input in cbreak/raw mode. The text input fix (`\r`) is correct for the same reason the MobileSessionView uses `\r`. The yesno buttons at lines 137 and 145 still send `"y\n"` and `"n\n"`. If Claude Code's interactive yesno prompts run in raw mode (as they do for text prompts), these will silently fail to confirm/deny. This is a pre-existing bug surfaced by this PR's fix, not introduced by it.

The corresponding test assertions (`"y\n"` / `"n\n"` at lines 304 and 325 of `NotificationDropdown.test.tsx`) are consistent with the unchanged source, so test coverage is correct as-is.

---

**On the specific review questions:**

1. **`\r` correctness for REST → PTY**: Confirmed correct. `handleSendInput` in the REST handler passes `req.Text` straight through to `callAgent("sendInput", ...)`, which RPC-dispatches to `Dispatcher.SendInput` in agentd, which calls `PTYManager.WriteInput` → `sess.master.Write(data)`. Bytes are written verbatim; `\r` is the right terminal Enter for raw-mode PTY.

2. **Other callers**: `SessionList.tsx` uses `"\r"` and `"\x1b"` (correct). `TicketCard.tsx` uses `"\r"` (correct). The only remaining `\n` callers are the yesno buttons noted above.

3. **Test changes**: Correct and complete — both "Send button" and "Enter key" test titles and assertions updated to `\r`.

4. **yesno regression concern**: No regression introduced by this PR. The yesno buttons were using `\n` before and still do; this PR did not change that. However, it does mean the buttons likely have the same bug the text input just had.

**Deferred:**
- yesno `"y\n"` / `"n\n"` → `"y\r"` / `"n\r"` fix → new ticket recommended, medium priority (same class of bug just fixed), discovered in #783 review

REVIEW STATUS: CHANGES NEEDED
- Fix yesno buttons to use `"y\r"` / `"n\r"` (same fix already applied to text input)

dokipen (commented):
Updated: also changed yesno buttons from `"y\n"`/`"n\n"` to `"y\r"`/`"n\r"` — same PTY path, same fix. All 786 tests pass. Both commits are ready for merge.

PR #784, opened 2026-04-26
fix(#675): skip ring buffer replay for TUI sessions, send SIGWINCH instead

dokipen:
## Summary
- Added `skipReplay bool` parameter to `PTYManager.ServeTerminal` — shell sessions replay the ring buffer on reconnect; TUI/agent sessions skip replay and send SIGWINCH instead
- SIGWINCH is sent by re-applying current PTY dimensions via `pty.Setsize`, which the kernel forwards as SIGWINCH to the foreground process group (no PID lookup needed; works for both created and reattached sessions)
- Added `ProfileType` field to `session.Session`, populated from `profile.Type` at creation and persisted across daemon restarts
- Added `IsShellSession(sessionID string) bool` to `SessionDispatcher` interface and implemented in `Dispatcher`; used in the relay path to determine `skipReplay` per session
- Added `TestServeTerminal_SkipReplay_TUISession` verifying the ring buffer is not delivered when `skipReplay=true`

Fixes #675

## Test plan
- [x] Verification checks pass (`shellcheck`)
- [x] All existing tests pass (`go test ./...`)
- [x] New test `TestServeTerminal_SkipReplay_TUISession` covers the skip-replay path
- [ ] Manual verification: reconnect xterm.js client to a running Claude or Gemini session — should show clean repainted screen (not garbled/blank)

### DIFF
--- a/services/agents/cmd/agentd/main.go
+++ b/services/agents/cmd/agentd/main.go
@@ -155,7 +155,11 @@ func main() {
 				slog.Warn("websocket accept failed", "error", err)
 				return
 			}
-			if err := ptyManager.ServeTerminal(r.Context(), sessionID, conn); err != nil {
+			skipReplay := true // default: TUI sessions skip ring buffer replay
+			if sess, lookupErr := manager.Get(sessionID); lookupErr == nil {
+				skipReplay = sess.ProfileType != "shell"
+			}
+			if err := ptyManager.ServeTerminal(r.Context(), sessionID, conn, skipReplay); err != nil {
 				slog.Warn("terminal session ended with error", "session_id", sessionID, "error", err)
 			}
 		})))
--- a/services/agents/internal/hub/client.go
+++ b/services/agents/internal/hub/client.go
@@ -33,6 +33,9 @@ type SessionDispatcher interface {
 	GetTerminalEndpoint(params json.RawMessage) (json.RawMessage, *rpcError)
 	GetDiagnostics(ctx context.Context, params json.RawMessage) (json.RawMessage, *rpcError)
 	SendInput(params json.RawMessage) (json.RawMessage, *rpcError)
+	// IsShellSession returns true if the session is a shell-type profile.
+	// Shell sessions replay the ring buffer on reconnect; agent/TUI sessions do not.
+	IsShellSession(sessionID string) bool
 }
 
 // Client manages the WebSocket connection from agentd to the hub.
@@ -316,8 +319,9 @@ func (c *Client) dispatchSessionAsync(ctx context.Context, conn *websocket.Conn,
 		if err := json.Unmarshal(resp.Result, &result); err == nil && result.Relay {
 			var p getTerminalEndpointParams
 			if err := json.Unmarshal(req.Params, &p); err == nil {
+				skipReplay := !c.dispatcher.IsShellSession(p.SessionID)
 				relayCtx, relayCancel := context.WithCancel(ctx)
-				go c.runTerminalRelay(relayCtx, relayCancel, conn, p.SessionID, c.ptyMgr)
+				go c.runTerminalRelay(relayCtx, relayCancel, conn, p.SessionID, c.ptyMgr, skipReplay)
 			}
 		}
 	}
--- a/services/agents/internal/hub/client_test.go
+++ b/services/agents/internal/hub/client_test.go
@@ -42,6 +42,7 @@ func (s *stubDispatcher) GetDiagnostics(_ context.Context, _ json.RawMessage) (j
 func (s *stubDispatcher) SendInput(_ json.RawMessage) (json.RawMessage, *rpcError) {
 	return json.RawMessage(`{}`), nil
 }
+func (s *stubDispatcher) IsShellSession(_ string) bool { return false }
 
 // TestClientReconnectsAfterSilentDisconnect verifies that the Client
 // reconnects when the TCP connection goes silent (no data, no close frame).
--- a/services/agents/internal/hub/dispatch.go
+++ b/services/agents/internal/hub/dispatch.go
@@ -418,6 +418,16 @@ type diagnosticsSummary struct {
 	ErrorCount         int `json:"error_count"`
 }
 
+// IsShellSession returns true if the session is a shell-type profile.
+// Returns false for agent/TUI sessions and on any lookup error.
+func (d *Dispatcher) IsShellSession(sessionID string) bool {
+	sess, err := d.manager.Get(sessionID)
+	if err != nil {
+		return false
+	}
+	return sess.ProfileType == "shell"
+}
+
 // SendInput handles the sendInput JSON-RPC method.
 // It writes the given text to the PTY master for the specified session.
 func (d *Dispatcher) SendInput(params json.RawMessage) (json.RawMessage, *rpcError) {
--- a/services/agents/internal/hub/relay.go
+++ b/services/agents/internal/hub/relay.go
@@ -75,6 +75,7 @@ func (c *Client) runTerminalRelay(
 	hubConn *websocket.Conn,
 	ptySessID string,
 	ptyMgr *pty.PTYManager,
+	skipReplay bool,
 ) {
 	parsed, err := uuid.Parse(ptySessID)
 	if err != nil {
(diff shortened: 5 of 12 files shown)


dokipen (commented):
## Code Review — PR #784: Skip ring buffer replay for TUI sessions

All tests pass. Build clean. Shellcheck passes.

### Summary

| Suggestion | `main.go:158-161` | `manager.Get` called outside the relay path duplicates the `IsShellSession` lookup logic | Consider consolidating via `IsShellSession` or a shared helper |
| Suggestion | `session/store.go:97` | `ProfileType` is an untyped `string`; values \"agent\" and \"shell\" are magic strings scattered across files | A typed constant (or small `ProfileType` type with consts) would catch typos at compile time |
| Suggestion | `pty/integration_test.go:172-181` | Test collects frames until 2-second context timeout — could be flaky on a very slow CI runner if live PTY output containing \"REPLAY_MARKER\" arrives after `skipReplay` (e.g., via SIGWINCH-triggered repaint) | Add a short explicit drain window or assert on the SIGWINCH repaint path separately |

### Detail

**`main.go` duplicate lookup (Suggestion)**

`main.go` reimplements the \"is shell?\" decision inline (`sess.ProfileType != "shell"`) rather than calling `d.dispatcher.IsShellSession`. The two paths will stay in sync as long as the string literal matches, but a future refactor changing the hub path would not automatically update the direct-WebSocket path. Low risk today; worth unifying.

**Untyped `ProfileType` string (Suggestion)**

`"shell"` and `"agent"` appear in `config.go`, `session/manager.go`, `session/store.go`, `hub/dispatch.go`, and `cmd/agentd/main.go`. A `type ProfileType string` with `const ProfileTypeAgent ProfileType = "agent"` etc. would make grep/refactor trivial and catch misspellings at compile time.

**Test drain window (Suggestion)**

`TestServeTerminal_SkipReplay_TUISession` runs the `sh -c "echo REPLAY_MARKER; sleep 30"` subprocess. After `Setsize`/SIGWINCH the shell itself won't repaint \"REPLAY_MARKER\" (it's not a TUI), so in practice this test is not flaky. The concern is theoretical — the test is correct as written.

### Atomicity / concurrency review

The snapshot-and-register block (`sess.mu.Lock` … `sess.mu.Unlock`) is unchanged in structure. `Setsize` is called *after* the writer is registered, so any repaint output triggered by SIGWINCH is guaranteed to be delivered to the new connection. No race.

### Persistence / restart safety

Sessions persisted before this PR have `profile_type` absent from JSON; `omitempty` means `recordToSession` sets `ProfileType = ""`. Both `IsShellSession` (returns `false`) and the `main.go` inline check (`"" != "shell"` → `skipReplay=true`) treat the empty value as TUI/agent — the safe default. Backward-compatible.

### Default for unknown sessions

In `main.go`, if `manager.Get` fails (session not yet in store), `skipReplay` remains `true` — the TUI-safe default. Correct.

---

REVIEW STATUS: APPROVED

All findings are suggestions; no critical or warning issues. The core logic — writer registration before SIGWINCH, persistence round-trip, safe defaults — is correct.

dokipen (commented):
## Test Review: `TestServeTerminal_SkipReplay_TUISession`

All tests pass cleanly. Full run output:

```
=== RUN   TestServeTerminal_SkipReplay_TUISession
--- PASS: TestServeTerminal_SkipReplay_TUISession (2.01s)
```

All other `TestServeTerminal` tests and the relay integration tests also pass with the updated `false` parameter at each call site.

---

### 1. Does the test adequately cover the `skipReplay=true` code path?

Partially. The test covers the primary negative assertion well: it confirms the ring buffer snapshot is not replayed to the client when `skipReplay=true`. That is the most important invariant for the TUI fix.

However, the other half of the `skipReplay` branch — the SIGWINCH delivery via `pty.Setsize` — is not verified at all. The test has no assertion that SIGWINCH was sent to the process. In practice, a regression that silently drops the `pty.Setsize` call (e.g., `GetsizeFull` failing) would go undetected. A TUI program that never receives SIGWINCH on reconnect will display a blank/stale screen.

**Gap: no SIGWINCH delivery test.** A test that confirms a SIGWINCH-responsive process (e.g., a small shell script using `trap`) actually repaints after `ServeTerminal(..., true)` is called would close this gap. That said, such a test would be more involved to write reliably and could reasonably be a follow-up.

### 2. Is the test correctly structured (no false positives/negatives)?

The structure is sound with one important caveat: **the test can produce a false negative** under a specific race.

The test polls until `REPLAY_MARKER` appears in the ring buffer, then connects with `skipReplay=true` and reads for 2 seconds. If the PTY process is still running and happens to emit more output after the WebSocket connects — which is unlikely given the `echo; sleep 30` command — that new live output could contain `REPLAY_MARKER` (it won't in this case, but the test comment says "Collect whatever the server sends"). The design is fine here because `sleep 30` emits nothing after the initial echo, so the test is deterministic in practice.

The 2-second `context.WithTimeout` for collecting output is adequate. The loop reads until timeout (which it always hits since `sleep 30` produces no further output), so `received` reliably contains only what arrives at connect time.

One structural note: the test does not assert that `received` is entirely empty. It only checks for the absence of `REPLAY_MARKER`. If the `skipReplay` branch sends any other data before the read loop starts (e.g., a future feature adds a reconnect banner), the test would still pass. This is intentional and appropriate for the bug being fixed.

### 3. Edge cases not covered

**a) SIGWINCH delivery verification (significant gap)**

The test does not confirm that `pty.GetsizeFull` + `pty.Setsize` actually fires and reaches the process. This is the "other half" of the `skipReplay=true` branch. A test could:
- Run a process with `trap 'echo SIGWINCH_RECEIVED' WINCH`
- Confirm no replay on connect
- Confirm `SIGWINCH_RECEIVED` appears in new output after connect

**b) `skipReplay=true` on reconnect during active PTY writes**

`TestServeTerminal_StrictModeReconnect_WritersContinuous` (relay layer) tests reconnect with `skipReplay=false`. There is no equivalent test for `skipReplay=true` during active writes. In the `skipReplay=true` path the snapshot is skipped but the writer is still registered before SIGWINCH is sent, so in-flight writes during SIGWINCH should be forwarded correctly. This is likely fine but untested at the pty layer.

**c) `skipReplay=true` on a session with empty ring buffer**

In `ServeTerminal`, when `skipReplay=false` and `len(snapshot) == 0`, the snapshot write is skipped. When `skipReplay=true`, SIGWINCH is always attempted regardless of buffer state. No test covers the `skipReplay=true` + empty buffer case, but the code path is simple and correct.

**d) `IsShellSession` false-negative on lookup failure**

`Dispatcher.IsShellSession` returns `false` (TUI behavior) when the session lookup fails. This means a lookup error causes replay to be skipped, which is the safer default for TUI users but could suppress replay for shell sessions if the session store has a transient error. This is a design tradeoff rather than a bug, and it is documented in the comment. No test covers this behavior.

### 4. Are the `false` parameter updates correct and complete?

Yes. All existing call sites in:
- `services/agents/internal/pty/integration_test.go`
- `services/agents/internal/hub/relay_integration_test.go`
- `services/agents/test/e2e/session_lifecycle_test.go`

have been updated to pass `false`, preserving the original shell-session replay semantics. The `stubDispatcher.IsShellSession` stub in `hub/client_test.go` correctly returns `false` (consistent with the default TUI behavior). The production call sites in `main.go` and `relay.go` use the new boolean from `IsShellSession`/`sess.ProfileType` as intended.

---

### Summary

The test covers the core skip-replay assertion correctly and the call-site updates are complete. The main gap is the absence of a SIGWINCH delivery assertion — the `skipReplay=true` branch does two things (skip replay + send SIGWINCH) but the test only verifies one of them. For the stated fix this is acceptable at merge time, but a follow-up test for SIGWINCH delivery would give full confidence that TUI reconnects actually trigger a repaint and not just suppress the garbled output.

dokipen (commented):
## Security Review

| Severity | Finding |
|----------|---------|
| ✅ Fixed | ** allow-listed on deserialization** —  now constrains to , resetting unknown values to  (safe TUI default). Pushed in follow-up commit. |
| Info | Session ID at HTTP endpoint is validated for path separators but not UUID format — lookup failure already defaults to  (safe) and  returns ; no exploitable risk, consistent with defense-in-depth already in place at the persist layer. |
| ✅ Safe | Fail-safe confirmed:  before lookup; only set to  on confirmed shell profile. |
| ✅ Safe | No IDOR via  — internal hub WebSocket only, keyed in-memory lookup. |
| ✅ Safe | No TOCTOU in / —  held for session lifetime;  is async-signal-safe.

dokipen (commented):
Security review complete. One fix applied: ProfileType is now allow-listed on deserialization in recordToSession (constrained to agent or shell; unknown values reset to agent). Confirmed safe: fail-safe defaults, no IDOR via IsShellSession, no TOCTOU in GetsizeFull/Setsize.

PR #785, opened 2026-07-31
fix(#677): stop agent disconnect loop from oversized prompt_context

dokipen:
## Summary

Fixes claude-cadence ticket #677 (issues-api): the mac agent flapped on/off the hub in a permanent reconnect loop — 16,281 occurrences of `StatusMessageTooBig / "RPC frame exceeds limit"` in its log.

Root cause: `PromptContext` is built as the "last 15 lines" of PTY scrollback, but TUI agents (claude, gemini) redraw with cursor-movement escapes rather than newlines, so those "15 lines" can span nearly the entire ~1 MiB ring buffer. The oversized `listSessions` response tripped the hub's 64 KiB `RPCMaxMessageSize`, the hub closed the connection, agentd reconnected, the hub called `listSessions` again — same frame, same close, forever. Restarting agentd never helped because the buffer refills as soon as the session idles at a prompt again.

Two changes, both needed:

- **agentd** (`services/agents/internal/session/monitor.go`): cap `PromptContext` at 8 KiB via `truncateTail`, keeping the tail (where the prompt is) and trimming to rune/line boundaries. This bounds the payload at the source — a worst-case context could exceed even the 1 MiB connection read limit, so no hub-side limit alone can fully close the bug.
- **hub** (`services/agent-hub/internal/hub/hub.go`): raise `RPCMaxMessageSize` 64 KiB → 512 KiB for headroom. Kept strictly below `MaxMessageSize` (1 MiB) so oversized text frames still reach the post-read check and produce the explicit close + log instead of dying opaquely at `SetReadLimit` (invariant enforced by `TestMaxMessageSizeConstants`).

## Test plan

- `go test ./...` — green in both `services/agent-hub` and `services/agents` (includes the oversized-frame close test and the `RPCMaxMessageSize < MaxMessageSize` invariant test)
- `make build build-linux` — clean for both services
- `shellcheck commands/**/*.sh skills/**/*.sh` — clean
- Manual QA pending: deploy hub to bootsy (`deploy-all.sh --skip-agents`), then agentd, and confirm `grep -c "RPC frame exceeds limit" agentd.log` stops incrementing (ticket #677 acceptance criteria)

N/A — no `services/issues-ui/` changes.

1. **Invariants**: `RPCMaxMessageSize < MaxMessageSize` (oversized text frames must reach the post-read check for the explicit close + structured log, rather than an opaque read-limit kill) — still enforced by `TestMaxMessageSizeConstants`. `PromptContext ≤ 8 KiB` at the producer, so `listSessions` responses stay far below the frame cap regardless of session age or TUI redraw behavior.
2. **If violated**: an agent with a long-lived TUI session gets disconnected on every `listSessions`, flapping offline permanently (the exact #677 failure).
3. **New trust boundaries / inputs**: none. The hub still enforces its limit on untrusted agent frames — the cap is loosened 8x but remains half the connection-level read limit; no unbounded input is newly accepted.

### DIFF
--- a/services/agent-hub/internal/hub/hub.go
+++ b/services/agent-hub/internal/hub/hub.go
@@ -30,11 +30,13 @@ const maxConsecutiveRPCFailures = 3
 const MaxMessageSize = 1 << 20
 
 // RPCMaxMessageSize is the maximum allowed size for a single JSON-RPC
-// (text) frame from an agent (64 KiB). RPC payloads (register, ping/pong,
-// listSessions, createSession responses) are structurally small — a 1 MiB
-// text frame would never occur in normal operation and likely indicates a
-// misconfigured or malicious client. Enforced post-read in HandleAgentConnection.
-const RPCMaxMessageSize = 64 * 1024
+// (text) frame from an agent (512 KiB). listSessions responses carry
+// prompt_context, which is derived from PTY scrollback: TUI agents redraw
+// with cursor movement rather than newlines, so "15 lines" of context can
+// span hundreds of KiB and a 64 KiB cap disconnected agents in a reconnect
+// loop. Must stay below MaxMessageSize so oversized text frames reach the
+// post-read check in HandleAgentConnection instead of dying at SetReadLimit.
+const RPCMaxMessageSize = 512 * 1024
 
 // Hub manages registered agentd connections.
 type Hub struct {
--- a/services/agents/internal/session/monitor.go
+++ b/services/agents/internal/session/monitor.go
@@ -195,9 +195,12 @@ func lastNLines(s string, n int) string {
 	return strings.Join(nonEmpty, "\n")
 }
 
-// truncateTail keeps the last max bytes of s, trimming a partial rune and a
-// partial leading line so the result starts on a line boundary.
+// truncateTail keeps the last max bytes of s, trimming a partial rune and,
+// when possible, a partial leading line.
 func truncateTail(s string, max int) string {
+	if max <= 0 {
+		return ""
+	}
 	if len(s) <= max {
 		return s
 	}
--- a/services/agents/internal/session/monitor.go
+++ b/services/agents/internal/session/monitor.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode/utf8"
 
 	"github.com/dokipen/claude-cadence/services/agents/internal/pty"
 )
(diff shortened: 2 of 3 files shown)


dokipen (commented):
## Code review (code-reviewer)

**Verdict: no blocking findings.** Fix is correct and well-targeted; both packages' tests pass.

- **Warning** — `truncateTail` panics on negative `max` (`s[len(s)-max:]`). Unreachable today (sole caller passes the 8192 constant) but needs a `if max <= 0 { return "" }` guard. → **being fixed in this PR**
- **Warning** — package convention is table-driven tests for pure helpers (`TestLastNLines`, `TestStripANSI`, …); `truncateTail` ships untested. → **being fixed in this PR**
- Suggestion — docstring overpromises "starts on a line boundary" (not guaranteed for no-newline TUI tails). → wording softened in the fix commit
- Suggestion — cut landing exactly on a line boundary discards one intact line (≤1 line of an 8 KiB window). Accepted as-is.
- UTF-8 handling verified correct (`utf8.RuneStart` skip, `IndexByte('\n')` safe against multibyte runes, exact-boundary no-op). 8 KiB cap confirmed compatible with `classifyPromptType` (end-anchored patterns see an intact tail) and the UI consumer.

dokipen (commented):
## Security review (security-engineer)

**Verdict: no Critical findings; nothing blocking merge.**

- **DoS surface of the 8x frame cap: no meaningful change.** Peak per-frame memory was always governed by the connection-level `SetReadLimit(MaxMessageSize)` (1 MiB) — the RPC check runs post-read, so 64 KiB→512 KiB does not change memory exposure. `/ws/agent` requires `HUB_AGENT_TOKEN` (constant-time compare) behind a rate limiter, so the delta applies only to authenticated agents. `TestMaxMessageSizeConstants` still guards the `RPC < MaxMessageSize` layering.
- **Data exposure: net improvement.** The 8 KiB cap strictly reduces the scrollback volume (potentially containing echoed secrets) leaving the host. Dashboard renders `prompt_context` as a React text child — no XSS path.
- **Warning (deferred → ticket #678)** — per-session cap ≠ per-frame cap: ~50–60 concurrently-waiting sessions can push one `listSessions` frame past 512 KiB and reproduce the #677 loop at scale. Tracked as a follow-up with acceptance criteria.
- **Warning** — negative-`max` panic in `truncateTail` (same as code review). → **being fixed in this PR**
- Suggestions (folded into the follow-up ticket): mid-line fragment can misclassify prompt type on no-newline tails; `ansiEscape` regex misses colon-subparameter CSI, DCS, bare ESC.

dokipen (commented):
Both review Warnings resolved in fa25ca0: negative-max guard + table-driven TestTruncateTail (8 cases incl. mid-rune cut, newline-final-byte, no-newline TUI tail; all outputs asserted valid UTF-8). Aggregate listSessions frame-size finding deferred to ticket #678. Review gate clear — proceeding to manual QA (agent-service paths touched).

PR #786, opened 2026-08-23
fix(#679): lowercase and trim normalizeRepo before comparison

dokipen:
## Summary

`normalizeRepo()` in `services/issues-ui/src/hooks/useAgents.ts` stripped SSH/HTTPS prefixes and the `.git` suffix from repo identifiers, but never lowercased, trimmed whitespace, or stripped a trailing slash before the strict `===` comparison used to match agent profiles to a project's repo. Case-mismatched repo slugs (e.g. `Owner/Repo` vs `owner/repo`) silently failed to match, causing Lead/Discuss/Refine launch actions to report "No online agents with profiles matching this repository" even when a correctly-scoped online agent existed.

The fix mirrors the existing correct pattern in `services/issues-cli/src/project-resolver.ts:80` (`.toLowerCase()` comparison) and the Go backend's `normalizeRepoFilter` in `services/agent-hub/internal/rest/handlers.go`, which already lowercases, strips `.git`, and strips a trailing slash.

`normalizeRepo` is shared by three call sites — `useAgents.ts:119` (`useAgentProfiles`), `AgentLaunchForm.tsx:32`, and `AgentManager.tsx:30` — so fixing the shared function fixes all three.

Fixes #679

## Test plan

- Added reproduction tests to `services/issues-ui/src/hooks/useAgents.test.ts` covering case-mismatch, trailing-slash, and whitespace-difference inputs; confirmed they failed before the fix and pass after.
- `npm run test -- src/hooks/useAgents.test.ts`: 25/25 passing (21 pre-existing + 4 new)
- `npm run test` (full issues-ui suite): 790/790 passing
- `npm run typecheck`: clean

N/A — pure string-normalization bug fix in a client-side repo-slug comparison used only to filter which agent profiles are displayed as launch options. No new trust boundaries or external inputs.

### DIFF
--- a/services/issues-ui/src/hooks/useAgents.test.ts
+++ b/services/issues-ui/src/hooks/useAgents.test.ts
@@ -62,12 +62,32 @@ describe("normalizeRepo", () => {
     expect(normalizeRepo("Owner/Repo")).toBe("owner/repo");
   });
 
-  it("treats a trailing-slash repo slug as equal to one without", () => {
-    expect(normalizeRepo("owner/repo/")).toBe(normalizeRepo("owner/repo"));
+  it("strips a trailing slash from a repo slug", () => {
+    expect(normalizeRepo("owner/repo/")).toBe("owner/repo");
   });
 
-  it("treats a whitespace-padded repo slug as equal to a trimmed one", () => {
-    expect(normalizeRepo("  owner/repo  ")).toBe(normalizeRepo("owner/repo"));
+  it("trims a whitespace-padded repo slug", () => {
+    expect(normalizeRepo("  owner/repo  ")).toBe("owner/repo");
+  });
+
+  it("handles combined case, whitespace, trailing slash, and .git suffix", () => {
+    expect(normalizeRepo("  Owner/Repo.GIT/  ")).toBe("owner/repo");
+  });
+
+  it("strips a trailing slash after a .git suffix", () => {
+    expect(normalizeRepo("https://github.com/owner/repo.git/")).toBe(
+      "owner/repo",
+    );
+  });
+
+  it("lowercases an uppercase SSH prefix", () => {
+    expect(normalizeRepo("GIT@GITHUB.COM:owner/repo")).toBe("owner/repo");
+  });
+
+  it("lowercases an uppercase HTTPS prefix", () => {
+    expect(normalizeRepo("HTTPS://GITHUB.COM/owner/repo")).toBe(
+      "owner/repo",
+    );
   });
 });
 
--- a/services/issues-ui/src/hooks/useAgents.test.ts
+++ b/services/issues-ui/src/hooks/useAgents.test.ts
@@ -53,6 +53,22 @@ describe("normalizeRepo", () => {
       "git@gitlab.com:owner/repo",
     );
   });
+
+  it("treats case-mismatched repo slugs as equal", () => {
+    expect(normalizeRepo("Owner/Repo")).toBe(normalizeRepo("owner/repo"));
+  });
+
+  it("lowercases a mixed-case repo slug", () => {
+    expect(normalizeRepo("Owner/Repo")).toBe("owner/repo");
+  });
+
+  it("treats a trailing-slash repo slug as equal to one without", () => {
+    expect(normalizeRepo("owner/repo/")).toBe(normalizeRepo("owner/repo"));
+  });
+
+  it("treats a whitespace-padded repo slug as equal to a trimmed one", () => {
+    expect(normalizeRepo("  owner/repo  ")).toBe(normalizeRepo("owner/repo"));
+  });
 });
 
 describe("normalizeRepo - nullish inputs from recovered sessions", () => {
--- a/services/issues-ui/src/hooks/useAgents.ts
+++ b/services/issues-ui/src/hooks/useAgents.ts
@@ -98,9 +98,12 @@ export interface AgentProfileEntry {
 export function normalizeRepo(repo: string | undefined): string {
   if (!repo) return "";
   return repo
+    .trim()
+    .toLowerCase()
+    .replace(/\.git$/, "")
+    .replace(/\/$/, "")
     .replace(/^git@github\.com:/, "") // SSH: git@github.com:owner/repo
-    .replace(/^https?:\/\/github\.com\//, "") // HTTPS/HTTP: https://github.com/owner/repo
-    .replace(/\.git$/, "");
+    .replace(/^https?:\/\/github\.com\//, ""); // HTTPS/HTTP: https://github.com/owner/repo
 }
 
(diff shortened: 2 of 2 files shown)


dokipen (commented):
Security review (low-risk UI-only change, no new trust boundary):

- No injection risk: `normalizeRepo` output is only used for client-side string comparison/filtering of agent profiles, never passed to a shell, DB query, or template.
- No ReDoS risk: all regexes are anchored (`^`/`$`) matching fixed literal strings, no nested quantifiers or backtracking-prone patterns.
- No security-relevant behavior change from case-insensitivity: since this only widens matching for *display filtering*, worst case is an over-inclusive match, not privilege escalation.

Minor correctness (non-security) finding:
- The `.git$` strip now runs *before* the trailing-slash strip and URL-prefix strip. For an input like `https://github.com/Owner/Repo.git/` (trailing slash after `.git`), the `.git` suffix is no longer stripped, producing `"owner/repo.git"` instead of `"owner/repo"` — a false-negative in matching. Low severity, edge case (unlikely stored form for a repo slug), but worth reordering (strip prefix → trailing slash → `.git`) or adding a test case if you want it airtight.

No blocking findings. Approving from a security standpoint modulo the above nit.

dokipen (commented):
## Test review — useAgents.test.ts (#679 fix)

Ran `npm run test -- src/hooks/useAgents.test.ts` in `services/issues-ui`: **25/25 passed**.

**Coverage of the fix:** The four new tests do exercise the three normalizations added to `normalizeRepo` (lowercasing, trailing-slash strip, trim), and the implementation now handles `.git` suffix stripping before the SSH/HTTPS prefix strips (order change also covered indirectly by pre-existing `.git`-suffix tests).

**Assertion quality — gap:** Three of the four new tests (`case-mismatched`, `trailing-slash`, `whitespace-padded`) only assert *equality between two `normalizeRepo()` calls* rather than asserting against a literal expected string. This can mask a bug where both sides are transformed identically but incorrectly (e.g., if a future edit made normalization return the input unchanged for both sides, these tests would still pass). Only the "lowercases a mixed-case repo slug" test checks a literal value (`toBe("owner/repo")`). Recommend adding direct literal assertions for the trailing-slash and whitespace cases too, e.g. `expect(normalizeRepo("owner/repo/")).toBe("owner/repo")` and `expect(normalizeRepo("  owner/repo  ")).toBe("owner/repo")`.

**Missing edge cases:**
- Combined case + whitespace + trailing-slash + `.git` in one input (e.g. `"  Owner/Repo.GIT/  "`) to confirm the full pipeline composes correctly.
- Uppercase/mixed-case SSH/HTTPS prefixes (e.g. `"GIT@GITHUB.COM:Owner/Repo"`, `"HTTPS://GITHUB.COM/Owner/Repo"`) — since lowercasing now happens *before* the prefix regexes (which are lowercase-only), these should now match after the fix, but there's no test locking that behavior in.
- Mixed-case `.GIT` suffix (e.g. `"owner/repo.GIT"`) — same reasoning; the `.git$` regex is case-sensitive and relies on the prior `.toLowerCase()` step, but nothing tests this ordering dependency directly.

**Severity:** Low/nice-to-have — the fix itself is verified working and the core regression (case mismatch) is covered. The gaps above are about hardening the test suite against future regressions in the normalization *order*, not blocking issues.

dokipen (commented):
## Code Review

| Warning | `services/issues-ui/src/hooks/useAgents.ts:100-105` | `.replace(/\.git$/, "")` runs *before* `.replace(/\/$/, "")`, so an input with both a `.git` suffix and a trailing slash (e.g. `git@github.com:Owner/Repo.git/` or `https://github.com/Owner/Repo.git/`) is normalized to `owner/repo.git` instead of `owner/repo` — the `.git` suffix survives. Verified: `normalizeRepo("git@github.com:Owner/Repo.git/")` → `"owner/repo.git"`. | Strip the trailing slash before (or after re-checking) the `.git` suffix, e.g. run `.replace(/\/$/, "")` first, then `.replace(/\.git$/, "")`, or apply the `.git` strip a second time after the slash strip. |
| Suggestion | `services/issues-ui/src/hooks/useAgents.ts:84-96` (JSDoc) | The doc comment still only describes prefix/suffix handling; it doesn't mention the new case-insensitivity, whitespace trimming, or trailing-slash stripping behavior. | Add a line noting normalization is case-insensitive and trims surrounding whitespace/trailing slash. |

### Detail

**Ordering bug (Warning):** The new `.trim().toLowerCase().replace(/\.git$/, "").replace(/\/$/, "")` chain strips `.git` before the trailing slash, so inputs combining both (plausible for pasted/copied clone URLs with an accidental trailing slash) aren't fully normalized. This is a real, if narrow, correctness gap directly relevant to the edge cases this PR is meant to cover. Low risk in practice but cheap to fix — recommend fixing now rather than deferring, e.g. reorder to strip the trailing slash first, or run the `.git$` replace once more after slash stripping.

Aside from that, the new tests are meaningful (case-mismatch, mixed-case, trailing slash, whitespace) and match the existing `describe`/`it` style in the file. The rest of the ordering (SSH/HTTPS prefix stripping happening after `toLowerCase()`) is correct since it makes the prefix regexes match regardless of input case.

Tests pass (`npx vitest run src/hooks/useAgents.test.ts`, 25/25).

REVIEW STATUS: CHANGES NEEDED
- Fix `.git` suffix vs. trailing-slash ordering bug in `normalizeRepo()`
- Update JSDoc to reflect new case/whitespace/trailing-slash normalization behavior

dokipen (commented):
Reviewed follow-up commit e5bed78d757fda8bf922e802683aa2b6089e46df on services/issues-ui/src/hooks/useAgents.ts and useAgents.test.ts.

**Ordering bug fix verified**: the `.replace()` chain now strips the trailing slash (`/\/$/`) before stripping `.git$`, so `"Owner/Repo.GIT/"` correctly normalizes to `owner/repo` instead of the previously-buggy `owner/repo.git`. Traced through all combinations (case + whitespace + trailing slash + `.git`, SSH/HTTPS with uppercase prefixes) and the reorder does not regress any other path — prefix stripping (SSH/HTTPS) still runs last and is unaffected by the slash/`.git` reorder.

**JSDoc updated**: now documents trim/lowercase/trailing-slash normalization before matching, addressing the prior suggestion.

**Tests**: strengthened with literal expected-value assertions (previously equality-between-two-calls, which could mask a shared bug), plus new coverage for the combined case+whitespace+trailing-slash+.git input and uppercase SSH/HTTPS prefixes. Ran `npm run test -- src/hooks/useAgents.test.ts` — all 29 tests pass.

No new issues introduced by the reorder. Both previously-reported items (ordering bug, JSDoc gap) are resolved.

REVIEW STATUS: APPROVED

PR #787, opened 2026-08-23
fix(#685): stop closing the agent connection on message size; bound payloads by construction

dokipen:
## Summary

The `mac` agentd flapped offline whenever a long-lived session was hovered in the sidebar. The hover opens the terminal relay, which replays the PTY ring buffer; a full buffer (1 MiB − 1) + ttyd prefix (1) + relay header (17) = **1,048,593 bytes — 17 over the hub's 1 MiB read limit** — so the websocket library closed the *entire* agent connection and every relayed terminal with it. Same class as #400 / #531 / #677: a size guard on the wire policing data that was never bounded by construction. This PR removes the mechanism rather than moving the threshold.

- **`services/shared/relay`**: `TtydFramePrefixLen`, `MaxPTYBufferSize`, `MaxSnapshotFrameSize`, and a 16 MiB `AgentMaxMessageSize` backstop so both modules compile against the same numbers; test asserts `MaxSnapshotFrameSize < AgentMaxMessageSize` with ≥4× margin.
- **`services/agent-hub`**: agent-link read limit is the shared backstop; `RPCMaxMessageSize` and the post-read size check that marked agents offline are **deleted** — no code path closes the connection or terminal channels because of message size any more. Browser terminal conns keep a separate 1 MiB `BrowserMaxMessageSize`; REST body caps use a REST-local `MaxSessionRequestBodySize`. Register ack advertises `max_message_bytes`. Relay-channel close made idempotent (the new snapshot test exposed a real double-close panic).
- **`services/agents`**: negotiates the limit from the register ack (512 KiB default for pre-#685 hubs); `prepareResponseFrame` bounds every RPC reply before writing — marshal → drop `prompt_context`/`prompt_type` → JSON-RPC `-32000` error — never sending a frame the hub would reject; marshal failures are no longer silently swallowed. `listSessions` (and `getDiagnostics`) are metadata-only; `getSession`/`createSession` keep the prompt fields. PTY buffer cap and `config.go` comment use the shared constants.
- **`services/issues-ui`**: `NotificationDropdown` fetches `prompt_context` per waiting session via new `fetchSession` / `useSessionPromptContext` (keyed on `idle_since`, only while the dropdown is open), with a "Loading prompt…" placeholder and graceful fallback.
- **docs**: runbook Mode C corrected — the trigger is the relay snapshot replay on hover/terminal-open, not persisted-state sync.

**Deploy note:** the relay fix is hub-side; deploying agent-hub alone stops the hover-triggered drops. The agentd half adds the negotiated limit, sender-side degrade, and metadata-only `listSessions`. Old agentd + new hub and new agentd + old hub both remain compatible (`max_message_bytes` is optional; agentd defaults to 512 KiB).

Fixes #685

## Test plan
- [x] Reproduction tests written first and confirmed failing for the right reason on `main` (`read limited at 1048577 bytes`; agent marked offline on 600 KiB text frame; constants invariant violated)
- [x] `shellcheck` verification passes
- [x] `go vet && go test ./...` green in `services/shared`, `services/agent-hub`, `services/agents`
- [x] `tsc -b` clean; vitest 43 files / 805 tests pass
- [x] New: `TestIntegration_MultiMiBListSessionsKeepsAgentOnline` (3 MiB `listSessions` passed through, agent online, pre-opened relay still delivers), `TestHandleAgentWebSocket_RegisterAdvertisesMaxMessageBytes`, `TestClientRegisterNegotiatesMaxMessageBytes`, `TestPrepareResponseFrame*`, `TestListSessions_WorstCasePayloadBounded` (200 sessions → 1.22 MB, < 4 MiB), `TestTerminalRelayCloseIsIdempotent`, `useSessionPromptContext.test.ts`
- [ ] Manual QA: deploy hub + agentd; hover/open a session with ≥1 MiB scrollback; host stays online and `last_seen` keeps advancing; notification dropdown still shows waiting prompts

### DIFF
--- a/docs/runbooks/agentd-mac-dead.md
+++ b/docs/runbooks/agentd-mac-dead.md
@@ -1,0 +1,200 @@
+# Runbook: `mac` agentd missing / "dead" in the dashboard
+
+**Symptom:** the `mac` host is missing from "available hosts" in the claude-cadence
+dashboard, or it blinks in and out.
+
+**Key fact:** agentd on `mac` talks to the hub over an **outbound** WebSocket — it
+dials the hub, the hub never dials agentd (see
+[`skills/agent-service/SKILL.md`](../../skills/agent-service/SKILL.md), "host loopback
+only"). So "mac is gone" almost never means the process crashed. It usually means the
+**registration with the hub is broken** while the process is still alive.
+
+There are three distinct failure modes. Triage tells you which one you have — **the fix
+differs**, and for one of them restarting does **not** help.
+
+---
+
+## Reference facts
+
+| Thing | Value |
+|---|---|
+| launchd label | `com.cadence.agentd` |
+| Binary | `/Users/bob/bin/agentd` |
+| Config | `~/.config/agentd/config.yaml` (operator-managed — do **not** read programmatically; it triggers macOS popups) |
+| Log | `/Users/bob/lib/agentd/agentd.log` |
+| Hub | `cadence.whatisbackdoor.com` → `192.168.86.28` (`bootsy.lan`), `:443` |
+| Terminal listener | `:8001` on the LAN IP (`advertise_address`) — LAN-bound by design, **not** localhost |
+
+---
+
+## Step 1 — Triage (always run this first)
+
+```bash
+# Is the process alive and managed by launchd?
+pgrep -lf '/Users/bob/bin/agentd'
+launchctl list | grep -i cadence
+
+# What does the hub actually think? (authoritative for the dashboard)
+curl -s -m 8 https://cadence.whatisbackdoor.com/api/v1/agents \
+  | jq -r '.agents[] | "\(.name)  status=\(.status)  last_seen=\(.last_seen)"'
+
+# What is agentd's own recent history?
+tail -30 /Users/bob/lib/agentd/agentd.log
+```
+
+Now match what you see to one of the three modes below.
+
+---
+
+## Mode A — Process is dead / launchd not running it
+
+**Signature:**
+- `pgrep` returns nothing, **or** `launchctl list | grep cadence` shows no
+  `com.cadence.agentd` line (or a non-zero exit status in the middle column).
+
+**Fix:**
+```bash
+launchctl kickstart -k gui/$(id -u)/com.cadence.agentd
+```
+If it immediately dies again, read the tail of the log for a startup/config error and
+escalate to the operator (config is operator-managed).
+
+---
+
+## Mode B — Zombie / stale WebSocket (process up, hub doesn't list mac)
+
+**Signature:**
+- `pgrep` shows a live agentd PID.
+- The hub agents list does **not** include `mac` (only `bootsy`, etc.).
+- agentd holds a TCP socket to the hub but its log shows **no recent**
+  `reconnecting to hub` / `connected to hub` activity — it has believed it was
+  "connected" for hours or days.
+
+```bash
+# Confirm: socket open at the OS level, but app-level registration is dead
+lsof -nP -a -p "$(pgrep -f '/Users/bob/bin/agentd')" -iTCP -sTCP:ESTABLISHED | grep ':443'
+```
+
+**Cause:** the hub on bootsy was restarted (wipes its in-memory agent registry), but
+mac's TCP socket stayed half-open, so agentd never noticed and never re-registered.
+
+**Fix:** restart agentd to force a fresh handshake.
+```bash
+launchctl kickstart -k gui/$(id -u)/com.cadence.agentd
+```
+
+**Verify:**
+```bash
+# Expect a fresh "connected to hub" line, then mac appears online
+tail -6 /Users/bob/lib/agentd/agentd.log
+curl -s -m 8 https://cadence.whatisbackdoor.com/api/v1/agents \
+  | jq -r '.agents[] | "\(.name)  status=\(.status)"'
+```
+
+> Restarting drops any sessions whose processes are already dead (logged as
+> `auto-destroying session: process no longer alive`) — that's cleanup, not lost work.
+> But it **will** interrupt genuinely live sessions, so check the session list first if
+> work is in flight.
+
+---
+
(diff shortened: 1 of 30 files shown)


dokipen (commented):
Reviewed against ticket #685 root-cause/plan comments and the documented AC adjustment. `go build`/`go test ./...` green in `services/agent-hub`, `services/agents`, `services/shared`; `shellcheck` clean.

| Warning (pre-existing, not introduced here) | `services/agent-hub/internal/hub/agent.go:DeliverTerminalFrame` | Relay lookup (`terminalMu.Lock/Unlock`) happens before the buffered `ch <- buf` send. Between unlock and send, `CloseTerminalChannel`/`CloseTerminalChannels`/`RegisterTerminalRelay` cleanup can close that same channel, causing a send-on-closed-channel panic. The new `terminalRelay{closeOnce}` fixes double-*close* races but does not fix send-after-close. Confirmed via `git show main:...agent.go` that this exact pattern predates this PR. | Track separately; capture `relay.ch` and a "closed" flag under `terminalMu`, or recover from the panic in the read loop. Given this PR's whole point is "message size never kills terminals," this race is adjacent and worth closing out. |
| Suggestion | `docs/plans/inline-prompt-answering.md`, `docs/inline-prompt-answering-techspec.md` | Stale references to `hub.RPCMaxMessageSize` and the old per-session `prompt_context` shape in `listSessions`/`getSession` split are historical planning docs and now inaccurate re: current wire shape. Low priority since they're not runbooks. | Leave as-is or add a note; not blocking. |

**Verified correct (no findings):**
- Hub Layer 1: `RPCMaxMessageSize` and the post-read close/offline path are fully removed (`hub.go`); `AgentMaxMessageSize` (16 MiB backstop) applied via `SetReadLimit` both in `handleAgentWebSocket` before the first read (register) and in `HandleAgentConnection` for the connection lifetime — same constant, no gap. REST body caps (`handleCreateSession`, `handleSendInput`) now use an independent `MaxSessionRequestBodySize` (512 KiB), decoupled from the hub RPC constant as required by AC.
- `register` ack advertises `max_message_bytes`; wire-level test (`TestHandleAgentWebSocket_RegisterAdvertisesMaxMessageBytes`) decodes raw JSON rather than the typed struct, catching tag drift.
- agentd `prepareResponseFrame`/`writeResponseFor`: marshal happens entirely before `writeMu.Lock()`; degrade path (shrink → re-marshal → JSON-RPC `-32000` error) is correct and never leaves `writeMu` held during marshal/degrade work. Error response `ID` is `string`-typed end-to-end, so it round-trips correctly. `defaultMaxMessageBytes = 512 KiB` matches the old hub's `RPCMaxMessageSize`, a sound backward-compat default. `shrinkableFor` covers `listSessions`/`getSession`/`createSession`, the only methods carrying `prompt_context`/`prompt_type`.
- `toSessionInfo(s, includePrompt)`: `listSessions`/`getDiagnostics` now correctly pass `false`; `getSession`/`createSession` pass `true`. Repo-wide grep for `prompt_context`/`promptContext` found no other list-response consumer relying on the field (issues-ui's only consumer is the new per-session `fetchSession`/`useSessionPromptContext` path; hub's `GET /api/v1/sessions` fan-out is a pure passthrough of agentd's now-metadata-only payload).
- UI: `useSessionPromptContext` correctly guards stale results via a `cancelled` closure flag, effect deps (`agentName, sessionId, enabled, idleSince`) are correct — no polling of its own. `NotificationDropdown` only mounts `NotificationItem` (and thus the per-session fetch hook) while the dropdown is `open`, so there's no N-session request storm on every poll tick — fetches only fire when a user opens the dropdown and only for waiting sessions. `promptType` string literals (`"yesno"`, `"select"`, `"text"`/`"shell"`) match backend `session/monitor.go` — no regression.
- Backward/forward compatibility: old agentd + new hub — hub omits `RPCMaxMessageSize` entirely so old agentd's writes are unaffected; old agentd doesn't send `max_message_bytes` awareness but that's a hub-side concern only. New agentd + old hub — `register()` gracefully falls back to `defaultMaxMessageBytes` (512 KiB) when `resp.Result` doesn't carry `max_message_bytes`, matching the old hub's actual enforced limit.
- No dead code, no duplicated size constants (all sourced from `services/shared/relay`), comments referencing the old `MaxMessageSize`/`RPCMaxMessageSize` semantics were all found only in explanatory "here's what used to happen" context, not stale documentation of current behavior.

REVIEW STATUS: APPROVED

dokipen (commented):
## Review round 1 — resolved

Four parallel reviews (code, security, performance, test quality). **No Critical findings.** Warnings and cheap suggestions fixed in `HEAD`; the rest deferred with tickets.

**Fixed in this PR**
- **Send-on-closed-channel race** in `ConnectedAgent.DeliverTerminalFrame` (pre-existing, flagged by code review): frames are now sent while holding `terminalMu`, and every close — including the registration `cleanup()` — happens under the same lock. `TestDeliverTerminalFrame_NoRaceWithClose` reproduces the race on the old code under `-race` and is clean now.
- `TestHandleAgentConnection_OversizedTextFrame` proves liveness via a real `ping` round-trip instead of a blind 6 s poll (~0.4 s now; verified it still fails if the old size-based close is reintroduced).
- `prepareResponseFrame` encode-failure case and `max_message_bytes` `0` / `-1` register cases added.

**Deferred (agent-discovered tickets)**
- #686 — pre-existing `-race` failure in `services/agents/internal/pty` (`TestServeTerminal_ResizeClamped`), reproduces on `main`.
- #687 — hardening: explicit read limit on agentd's hub connection; optional cap on concurrent agent connections at the hub.

**Security assessment:** the 16 MiB agent-link backstop is acceptable — `HUB_AGENT_TOKEN` is verified before upgrade/read, no browser-reachable path inherits it, REST caps are intact, and no new logs include prompt or terminal bytes.
**Performance:** common RPC path is unchanged (marshal + one `len` check); per-session prompt fetches only fire while the dropdown is open and re-fire only on `idle_since` change.

Not addressed (historical planning docs under `docs/plans/` still mention `RPCMaxMessageSize`; left as-is intentionally).

PR #789, opened 2026-08-25
fix(#692): reset hub reconnect backoff after a stable connection

dokipen:
## Summary

- `connectLoop` in `services/agents/internal/hub/client.go` maintained an `attempt` counter for exponential reconnect backoff that was supposed to reset to 0 after a successful, stable connection, but `connect()` never actually returns `nil` (its only exit path is `readLoop` wrapping a `conn.Read` error) — so the `else { attempt = 0 }` branch was dead code and `attempt` accumulated for the life of the process, causing ~30s reconnect waits even after days of healthy uptime.
- Fixed by timing each connection attempt: when `connect()` fails, `attempt` resets to 0 if the connection stayed up for at least one `ReconnectInterval` (treated as "stable"); otherwise it increments as before, so rapid connect/disconnect loops still back off.
- Added a regression test, `TestClientResetsBackoffAfterStableConnection`, reproducing the accumulated-backoff behavior.

Fixes #692

## Test plan

- [x] `go test -v ./internal/hub/...` in `services/agents` — all tests pass, including the new regression test and existing `TestClientReconnectsAfterSilentDisconnect`, `TestClientNoKeepaliveWhenDisabled`, `TestBackoff`
- [x] `go build ./...` and `go vet ./...` clean
- [x] `shellcheck commands/**/*.sh skills/**/*.sh` (project verification command) — clean, no shell scripts touched

This PR touches reconnection logic in the agentd hub client.

1. **Invariant protected:** the backoff counter must only reset after a connection has demonstrated real stability (stayed up >= one `ReconnectInterval`), never on every disconnect — otherwise a tight connect/drop loop (e.g. a flapping network path or a hub rejecting the agent) would retry near-instantly forever instead of backing off, which could hammer the hub.
2. **What breaks if violated:** resetting too eagerly reintroduces a thundering-herd/tight-loop risk against the hub; not resetting at all is the original bug (agents appear offline for 30+s on every reconnect after a healthy run).
3. **New trust boundaries/inputs:** none — this only changes internal timing logic based on `time.Now()`/`ReconnectInterval` (existing config), no new external input is introduced.

### DIFF
--- a/services/agents/internal/hub/client.go
+++ b/services/agents/internal/hub/client.go
@@ -121,16 +121,21 @@ func (c *Client) connectLoop(ctx context.Context) {
 		default:
 		}
 
+		connStart := time.Now()
 		err := c.connect(ctx)
 		if err != nil {
 			if ctx.Err() != nil {
 				return
 			}
-			slog.Warn("hub connection failed", "error", err, "attempt", attempt)
-			attempt++
-		} else {
-			// Reset backoff after a successful connection that later disconnected.
-			attempt = 0
+			if time.Since(connStart) >= c.cfg.ReconnectInterval {
+				// The connection stayed up long enough to be considered stable;
+				// treat this disconnect as fresh and reset the backoff.
+				slog.Warn("hub connection failed after stable period, resetting backoff", "error", err)
+				attempt = 0
+			} else {
+				slog.Warn("hub connection failed", "error", err, "attempt", attempt)
+				attempt++
+			}
 		}
 
 		// Exponential backoff with jitter: 1s → 30s max.
--- a/services/agents/internal/hub/client.go
+++ b/services/agents/internal/hub/client.go
@@ -127,10 +127,17 @@ func (c *Client) connectLoop(ctx context.Context) {
 			if ctx.Err() != nil {
 				return
 			}
-			if time.Since(connStart) >= c.cfg.ReconnectInterval {
+			stableThreshold := c.cfg.ReconnectInterval
+			if stableThreshold <= 0 {
+				// Match backoff()'s floor: an unset/misconfigured
+				// ReconnectInterval must not make every disconnect look
+				// "stable" and disable backoff entirely.
+				stableThreshold = time.Second
+			}
+			if time.Since(connStart) >= stableThreshold {
 				// The connection stayed up long enough to be considered stable;
 				// treat this disconnect as fresh and reset the backoff.
-				slog.Warn("hub connection failed after stable period, resetting backoff", "error", err)
+				slog.Warn("hub connection failed after stable period, resetting backoff", "error", err, "attempt", attempt)
 				attempt = 0
 			} else {
 				slog.Warn("hub connection failed", "error", err, "attempt", attempt)
--- a/services/agents/internal/hub/client_test.go
+++ b/services/agents/internal/hub/client_test.go
@@ -236,6 +236,8 @@ func TestClientResetsBackoffAfterStableConnection(t *testing.T) {
 
 	var connCount atomic.Int64
 	var mu sync.Mutex
+	var acceptTime1 time.Time
+	var acceptTime2 time.Time
 	var closeTime3 time.Time
 	var acceptTime4 time.Time
 	conn4Seen := make(chan struct{})
@@ -245,6 +247,18 @@ func TestClientResetsBackoffAfterStableConnection(t *testing.T) {
 
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		n := connCount.Add(1)
+
+		if n == 1 {
+			mu.Lock()
+			acceptTime1 = time.Now()
+			mu.Unlock()
+		}
+
+		if n == 2 {
+			mu.Lock()
+			acceptTime2 = time.Now()
+			mu.Unlock()
+		}
 
 		if n == 4 {
 			mu.Lock()
@@ -318,15 +332,38 @@ func TestClientResetsBackoffAfterStableConnection(t *testing.T) {
 	client.Start()
 	t.Cleanup(client.Stop)
 
+	// Widened from 5s: flakiness runs (~130 iterations, some with -race)
+	// showed occasional stalls of ~5.3-5.4s before this point, well within
+	// the old 5s bound's margin of error. The stall happens in the initial
+	// dial/registration round-trip, before the timed portion below, so
+	// widening this doesn't affect what the test measures.
 	select {
 	case <-conn4Seen:
-	case <-time.After(5 * time.Second):
+	case <-time.After(25 * time.Second):
 		t.Fatalf("timed out waiting for 4th connection attempt, saw %d connection(s)", connCount.Load())
 	}
 
 	mu.Lock()
+	gap12 := acceptTime2.Sub(acceptTime1)
 	gap := acceptTime4.Sub(closeTime3)
 	mu.Unlock()
+
+	// Acceptance criterion (b): rapid connect/disconnect cycles still back
+	// off. Connection #1 fails immediately, so the delay before connection
+	// #2 is dialed reflects backoff(attempt=1, reconnectInterval) -- roughly
+	// 2x the base interval -- not backoff(attempt=0, reconnectInterval),
+	// which would be roughly 1x. With ±25% jitter, attempt=0 tops out at
+	// 1.25x base while attempt=1 bottoms out at 1.5x base, so a threshold of
+	// 1.4x base cleanly distinguishes the two. A regression that reset
+	// attempt unconditionally on every failure (rather than only after a
+	// stable connection) would collapse this gap back into the attempt=0
+	// range, so assert it's meaningfully larger than that.
+	minWant12 := time.Duration(1.4 * float64(reconnectInterval))
+	if gap12 < minWant12 {
+		t.Fatalf("reconnect delay before 2nd connection attempt = %v, want >= %v (roughly 2x base reconnect interval with jitter); "+
+			"this indicates a failed connection attempt is not being backed off (issue #692 acceptance criterion b)",
+			gap12, minWant12)
+	}
 
 	// A correctly-reset attempt counter yields a delay close to the base
 	// reconnectInterval (up to ±25% jitter, i.e. <= 1.25x). Give a generous
--- a/services/agents/internal/hub/client_test.go
+++ b/services/agents/internal/hub/client_test.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


dokipen (commented):
Performance review of the reconnect-backoff fix (`connectLoop` in `services/agents/internal/hub/client.go`).

**Suggestion (no blocking issues found):**

- `time.Since(connStart)` is a single monotonic-clock read per loop iteration (each iteration already does a network dial + WS handshake, so this is negligible overhead — no concern here).
- No tight-loop / thundering-herd risk introduced: a connection that fails *before* one `ReconnectInterval` elapses still increments `attempt` and goes through the existing exponential-backoff-with-jitter path (`backoff()`), so rapid flap/reject scenarios (e.g. hub rejecting auth) still back off up to the 30s cap exactly as before. The reset only fires when the connection actually stayed up for a full `ReconnectInterval`, which is a reasonable, cheap-to-compute proxy for "stability" and matches the intent described in the PR's invariant section.
- Threshold choice (`>= 1x ReconnectInterval`, typically on the order of a few seconds per `config_test.go` defaults) seems reasonable as a stability bar — it's short enough to recover quickly from a real healthy-then-dropped connection, and long enough that a connection which registers successfully but immediately drops (e.g. hub-side rejection right after ack) won't falsely reset backoff and hammer the hub. Worth double-checking in practice whether an unusually low configured `ReconnectInterval` (sub-second) could let a flapping-but-technically-"stable" connection reset backoff too eagerly under real hub load with many agents reconnecting simultaneously — but this is a config/tuning concern rather than a bug in the change itself.
- No resource leak concerns: `connStart` is a local per-iteration variable, and the existing `conn.Close`/`closeConn` paths are untouched by this diff.

Overall: low risk, appropriately scoped fix. No Critical or Warning-level performance findings.

dokipen (commented):
Reviewed the diff and test in `services/agents/internal/hub/client.go` / `client_test.go`. The stability-window approach (reset backoff only if the connection stayed up >= one `ReconnectInterval`) correctly fixes the dead-code bug in #692 while still backing off on tight connect/drop loops.

| Warning | `client.go:132` (`time.Since(connStart) >= c.cfg.ReconnectInterval`) | If `hub.reconnect_interval` is misconfigured to `0` (or negative), the stability check is always true (`time.Since >= 0`), so `attempt` resets to 0 on *every* failure, including instant connect/drop loops. `backoff()` already guards `base <= 0` by falling back to `1s`, but this new check doesn't share that guard, silently defeating the anti-thundering-herd protection this PR is meant to preserve. | Reuse the same floor as `backoff()` (e.g. compute `base := c.cfg.ReconnectInterval; if base <= 0 { base = time.Second }` once, and use it both for the stability comparison and the `backoff()` call) so a `0`/negative config can't disable backoff entirely. |
| Suggestion | `client_test.go:336` (`maxWant := 2 * reconnectInterval` = 60ms) | The margin between `closeTime3` and `acceptTime4` is fairly tight (60ms) compared to other timing-based tests in this file (which use ±25% margins on multi-hundred-ms/second scales); on a loaded CI runner, TCP dial + WS handshake + register could plausibly exceed this, causing an intermittent flake. | Consider widening the margin (e.g. 3-4x `reconnectInterval`) or bumping `reconnectInterval` slightly for more headroom, consistent with the ± tolerances used elsewhere in this file. |

**Suggestions:** the reset-branch log line drops the `attempt` field that the increment-branch log includes, which loses useful context (how many attempts were reset) — consider logging the prior `attempt` value on the reset path for symmetry.

No correctness issues found otherwise; logging is clear, no race conditions in `connStart`/`time.Since` usage (single goroutine), and the change is consistent with the file's existing style. The regression test itself correctly reproduces the original bug (fails without the fix, passes with it).

**Deferred:** none — the zero/negative `ReconnectInterval` guard is cheap to fix in this PR rather than deferring.

REVIEW STATUS: CHANGES NEEDED
- Add a floor guard (matching `backoff()`'s `base <= 0` handling) to the stability check so a misconfigured `reconnect_interval: 0` doesn't disable backoff entirely.

dokipen (commented):
Security review of the backoff-reset fix in `connectLoop` (`services/agents/internal/hub/client.go`).

## Findings

| Suggestion | client.go:124-138 | `connStart` is set immediately before `c.connect(ctx)`, so the "stability window" includes dial + TLS + register RPC latency, not just genuine post-registration uptime. A hub (or an intermediary that terminates/proxies the WS handshake) that reliably takes ~`ReconnectInterval` to complete the handshake and then fails immediately after registering can keep `attempt` pinned at 0 forever — the exponential backoff never engages even though every connection is failing right after setup. | Start the stability timer after `register()` succeeds (i.e. measure uptime of the *established* session), not from dial start, so backoff still escalates for a hub that consistently fails immediately post-handshake. |
| Suggestion | client.go:130 | `time.Since(connStart) >= c.cfg.ReconnectInterval` — the reset only re-arms the base delay, it never lets retries go *faster* than `ReconnectInterval`, so a malicious/misbehaving hub cannot induce a tighter reconnect loop than the operator-configured base interval; worst case an adversarial hub can suppress backoff escalation and hold the agent at steady-state `ReconnectInterval` retries indefinitely. Low impact since the hub already fully controls accept/close timing for connections it terminates, and a MITM capable of completing the authenticated register handshake already has a much stronger position (token/traffic access) than this gains it. No fix required, noting for the record. | None required — document as accepted risk. |
| Info | client.go:161-166 | Pre-existing (not introduced by this diff): `websocket.Dial` uses the long-lived loop `ctx` with no per-attempt dial timeout, so a hub that accepts the TCP connection but never completes the WS upgrade can hang `connect()` indefinitely; `attempt` never increments and no backoff occurs during the hang. Not exploitable by this PR's changes but adjacent to the reconnect-robustness goal. | Consider a per-dial timeout in a follow-up; out of scope for #692. |

## Assessment of the requested threat scenarios

- **Gaming the stability threshold to hammer the hub:** Not really possible beyond steady-state `ReconnectInterval` — see Suggestion #2. The reset logic never produces retries faster than the base interval, so there's no thundering-herd/DoS amplification vs. today's already-legitimate "stable connection dropped" case.
- **Malicious hub tricking a well-behaved agent into resetting backoff prematurely:** Possible via the dial/handshake-time inclusion (Suggestion #1), but the practical effect is "agent retries at base interval instead of escalating," which is bounded and not more aggressive than the intended steady-state behavior for a genuinely stable-then-dropped connection. Impact is limited to wasted agent/hub cycles at a rate the operator already configured as acceptable.
- **Other auth/connection handling in the diff:** No new external inputs are introduced (confirmed against the diff — only `time.Now()`/`ReconnectInterval` timing logic changed). Token handling (`ResolveToken()`, `Authorization: Bearer` header) is untouched by this PR.

No Critical/High/Medium findings. The two Suggestions are non-blocking; recommend addressing Suggestion #1 (measure stability from post-registration, not dial-start) as a quick follow-up since it's a small change that meaningfully tightens the invariant described in the PR's own "Attack Surface" section, but it does not need to block merge.

dokipen (commented):
## Test review: `TestClientResetsBackoffAfterStableConnection` (services/agents/internal/hub/client_test.go)

Ran the new test 130+ times total (`-count=10/20/30/40`, with and without `-race`) to check flakiness, plus added temporary instrumentation (reverted before this comment) to inspect the measured `gap` and connection counts across runs.

**Flakiness result:** 0 failures across ~130 runs. Typical duration ~0.33–0.41s. However, in a handful of runs (roughly 1–7 out of every 30, worse under `-race` / system load, rarer under light load) the test took ~5.3–5.4s instead — landing suspiciously close to the test's own hard-coded 5s timeout on `conn4Seen`:
```go
select {
case <-conn4Seen:
case <-time.After(5 * time.Second):
    t.Fatalf("timed out waiting for 4th connection attempt, saw %d connection(s)", connCount.Load())
}
```
Instrumenting `gap` showed these slow runs still produced a small `gap` (tens of ms) once `conn4Seen` finally fired — i.e., the multi-second stall happens *before* connection 3/4 (most likely during the very first dial/registration round-trip, possibly first-run warmup of the shared `http.DefaultClient`/transport that `websocket.Dial` uses implicitly, since `connect()` never sets a custom `HTTPClient` in `DialOptions`). It never manifested as a `gap` assertion failure in my runs, but it did come within ~400ms of blowing through the test's own 5s timeout. On a slower/busier CI runner this is a real risk of an outright `t.Fatalf` unrelated to the backoff logic being tested.

**Severity: Warning** — Consider bumping the `conn4Seen` wait timeout (e.g. to 10–15s) to give headroom, or explicitly warming up a dedicated `http.Client`/dialer before starting the timed portion of the test, since the failure mode here is a spurious timeout, not a wrong assertion.

---

### Coverage: acceptance criterion (b) is not actually asserted

The ticket's fix has two behaviors to verify:
(a) attempt resets after a stable connection, and
(b) rapid connect/disconnect cycles still back off (not reset eagerly).

The test builds up `attempt` via two immediate failures (conns #1, #2) before the "stable" connection #3, and the PR description leans on this to imply (b) is covered. But the test **only asserts** the gap between conn #3's close and conn #4's accept is small. It never asserts anything about the delay between conn #1→#2 or #2→#3 (i.e., it never checks that those delays actually grow/back off).

Concretely: if a regression made the loop reset `attempt = 0` on *every* failure (not just stable ones — i.e., removed the `time.Since(connStart) >= ReconnectInterval` guard entirely), conns #1 and #2 would incorrectly reset backoff instantly, but the test would still pass, because attempt would be reset to 0 again after conn #3 regardless. This test does not actually exercise/guard criterion (b).

**Severity: Warning** — Recommend adding an explicit assertion that the delay between conn #1→#2 (a genuinely rapid failure pair, both under `ReconnectInterval`) grows relative to the base interval (e.g., record timestamps for connections 1 and 2 and assert the resulting delay is meaningfully larger than the jittered base, analogous to how `TestBackoff` already checks exponential growth) — otherwise criterion (b) has no regression coverage at the `connectLoop` integration level.

---

### Resource / goroutine cleanup

`t.Cleanup` usage (`close(connHold)`, `srv.Close`, `client.Stop`) follows the existing pattern in this file (matches `TestClientReconnectsAfterSilentDisconnect`) and cleanup runs in LIFO order (`client.Stop()` → `srv.Close()` → `close(connHold)`), so the client is stopped and readLoop/keepalive goroutines torn down before the server handler goroutines are released via `connHold`. No leaked goroutines observed with `-race` across 70+ runs.

**Severity: Suggestion** — Minor: the `n == 4` handler records `acceptTime4` *before* `websocket.Accept` completes the WS handshake, so `gap` includes handshake latency on top of the actual reconnect-scheduling delay. This is fine given the current 2x tolerance margin but worth a comment noting it's intentional (measuring "time until the hub sees the next TCP-level attempt," not "time until registration completes").

---

### Overall

The core fix in `client.go` (timing each connection via `connStart` and resetting `attempt` only when the connection was up for at least one `ReconnectInterval`) looks correct and directly addresses the dead-code bug described in the PR summary (the old `else` branch on `connect() == nil` was indeed unreachable, since `connect()`'s only return path is via `readLoop` returning a non-nil error). The regression test is a good-faith reproduction of criterion (a) and is not coincidentally passing for that criterion — it would fail under the pre-fix code, since `attempt` would still be 3 (not reset) going into conn #4's backoff calculation. The two Warnings above (timeout headroom, missing assertion for criterion (b)) are the main gaps worth addressing before merge; neither blocks the correctness of the underlying fix.

dokipen:
Addressed both review notes in `connectLoop` (services/agents/internal/hub/client.go):

1. **Floored the stability threshold**: `c.cfg.ReconnectInterval` now falls back to `1s` when `<= 0`, mirroring the existing floor in `backoff()`, so a misconfigured/zero interval no longer trivially satisfies `time.Since(connStart) >= threshold` and defeats backoff on every failure.
2. **Added the missing `attempt` field** to the reset-branch log line (`hub connection failed after stable period, resetting backoff`), matching the increment-branch log.

Verified with `go build ./...`, `go vet ./...`, and `go test -v ./internal/hub/...` — all clean, all existing tests passing.

dokipen:
Addressed both review findings in `TestClientResetsBackoffAfterStableConnection` (services/agents/internal/hub/client_test.go):

1. **Timeout flakiness**: Widened the wait for `conn4Seen` from 5s to 25s. The observed ~5.3-5.4s stalls happen during the first dial/registration round-trip, before the measured `gap` window, so this only adds setup slack and doesn't change what's being measured.

2. **Missing acceptance-criterion assertion (rapid connect/disconnect cycles still back off)**: Extended the `Accept` handler to timestamp connections #1 and #2, then assert the gap between them is >= 1.4x `reconnectInterval`. With ±25% jitter, `backoff(attempt=0, base)` tops out at 1.25x base while `backoff(attempt=1, base)` bottoms out at 1.5x base, so 1.4x cleanly distinguishes "still backing off after an immediate failure" from "attempt was reset". Verified by reasoning through `connectLoop`'s attempt/stableThreshold logic: a regression that reset `attempt` unconditionally on every failure (rather than only after a connection exceeding the stable threshold) would collapse this gap into the attempt=0 range and fail the new assertion.

Ran `-run TestClientResetsBackoffAfterStableConnection -count=20 -v` and `-count=10 -race`: all pass. `go vet ./...` is clean.

PR #790, opened 2026-08-25
fix(#691): close websocket connection when marking agent offline for liveness

dokipen:
## Summary

- `markOfflineIfCurrent` (`services/agent-hub/internal/hub/hub.go`) now closes the agent's websocket connection (`websocket.StatusGoingAway, "liveness check failed"`) inside the existing `current == expected` guard, alongside setting status offline. This single change covers all three liveness paths that call it: `heartbeatLoop` (ping write failure and pong timeout), `wsKeepaliveLoop` (protocol-level ping failure), and `Call`'s RPC-failure demotion.
- The reaper now closes a stale agent's connection (`websocket.StatusGoingAway, "reaped: stale agent"`) before evicting it from the map, fixing the same deadlock for the sleep/wake + reaper scenario described in the ticket.
- Without this, `HandleAgentConnection`'s read loop stayed alive on a connection whose agent was marked offline, kept calling `Touch()`, and the websocket library auto-answered agentd's protocol pings — so agentd's own keepalive kept succeeding and it never reconnected. One transient heartbeat hiccup (or any laptop sleep/wake cycle) permanently demoted the agent until the TCP connection happened to drop.
- Grepped the repo for other callers of the exported `MarkOffline` — none in production code, only tests, so no other call sites needed the fix.

Fixes #691

## Test plan

- Reproduced the bug first with a test proving the connection stayed open and `Touch()` kept firing after an offline transition (this test is now inverted into `TestHeartbeatTimeoutClosesConnection`, asserting the connection is actually closed).
- Added `TestHeartbeatTimeoutThenReconnectRestoresOnline`, covering the acceptance criterion end-to-end: heartbeat timeout → offline → connection closed → agent reconnects/re-registers → hub reports it online again, with no manual intervention.
- `cd services/agent-hub && go build ./... && go vet ./...` — clean
- `gofmt -l internal/hub/hub.go internal/hub/hub_test.go` — clean (pre-existing gofmt issues in `internal/config/config_test.go`, `internal/proxy/terminal_test.go`, `internal/rest/server.go` are unrelated to this change and untouched by this PR)
- `go test ./internal/hub/... -v -count=1` — full package suite passes
- `go test ./internal/hub/... -count=5` — no flakiness across 5 runs
- `go test ./internal/hub/... -race -count=1` — no races detected

This PR touches connection/reconnection lifecycle logic in agent-hub.

1. **Invariant protected**: an agent's `status` and its underlying connection's liveness must stay in sync — if status is `offline`, the connection must not still be readable/writable, and if a connection is still open and serving reads, the status must not be `offline`. Previously this invariant was violated: status could be `offline` while the connection stayed fully alive.
2. **What breaks if violated**: the hub and the real agentd process disagree about connection state indefinitely (the deadlock this ticket describes) — the UI reports the agent unreachable while it's actually healthy and connected, until something external (TCP drop, process restart) breaks the tie.
3. **New trust boundaries / external inputs**: none. No new external inputs are introduced; this only changes internal handling of existing liveness-failure and reaper code paths. The `Close()` calls follow the same pattern already used by `Stop()` and `Register()`'s replace-on-reconnect path, and are nil-guarded for the (test-only) case of agents registered with a nil connection.

### DIFF
--- a/services/agent-hub/internal/hub/hub.go
+++ b/services/agent-hub/internal/hub/hub.go
@@ -508,16 +508,24 @@ func (h *Hub) reaper(ctx context.Context) {
 		case <-ticker.C:
 			h.mu.Lock()
 			now := time.Now()
+			var staleConns []*websocket.Conn
 			for name, agent := range h.agents {
 				if agent.Status() == StatusOffline && now.Sub(agent.LastSeen()) > h.agentTTL {
 					slog.Info("reaping stale agent", "agent", name, "last_seen", agent.LastSeen())
 					if conn := agent.Conn(); conn != nil {
-						conn.Close(websocket.StatusGoingAway, "reaped: stale agent")
+						staleConns = append(staleConns, conn)
 					}
 					delete(h.agents, name)
 				}
 			}
 			h.mu.Unlock()
+
+			// Close connections outside the lock: Close can block for the
+			// duration of the close handshake (up to several seconds), and
+			// must not stall other hub map operations while reaping.
+			for _, conn := range staleConns {
+				conn.Close(websocket.StatusGoingAway, "reaped: stale agent")
+			}
 		}
 	}
 }
--- a/services/agent-hub/internal/hub/hub.go
+++ b/services/agent-hub/internal/hub/hub.go
@@ -210,6 +210,9 @@ func (h *Hub) markOfflineIfCurrent(name string, expected *ConnectedAgent) {
 
 	if ok && current == expected {
 		expected.SetStatus(StatusOffline)
+		if conn := expected.Conn(); conn != nil {
+			conn.Close(websocket.StatusGoingAway, "liveness check failed")
+		}
 		slog.Info("agent marked offline", "agent", name)
 	}
 }
@@ -508,6 +511,9 @@ func (h *Hub) reaper(ctx context.Context) {
 			for name, agent := range h.agents {
 				if agent.Status() == StatusOffline && now.Sub(agent.LastSeen()) > h.agentTTL {
 					slog.Info("reaping stale agent", "agent", name, "last_seen", agent.LastSeen())
+					if conn := agent.Conn(); conn != nil {
+						conn.Close(websocket.StatusGoingAway, "reaped: stale agent")
+					}
 					delete(h.agents, name)
 				}
 			}
--- a/services/agent-hub/internal/hub/hub_test.go
+++ b/services/agent-hub/internal/hub/hub_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"strings"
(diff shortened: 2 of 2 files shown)


dokipen (commented):
## Security review

Scope: `services/agent-hub/internal/hub/hub.go` (+6 lines) and matching tests. Reviewed for new attack surface, DoS/resource-exhaustion potential, and trust-boundary changes around agent close/reconnect.

| Medium | `services/agent-hub/internal/hub/hub.go:509-520` (reaper) | `conn.Close()` is now called once per stale agent *while holding the hub-wide write lock* (`h.mu.Lock()`). `coder/websocket`'s `Close()` performs a synchronous close handshake with an internal timeout, so if several agents go offline+stale in the same tick (or a peer is slow/unresponsive to the close handshake), the reaper serially blocks inside the critical section, stalling `Register`, `Get`, `List`, and `MarkOffline` (all lock/RLock-gated) for the cumulative handshake time. This is new behavior introduced by this PR — the reaper never touched the network under the lock before. | Move `conn.Close()` calls outside the locked section: collect `(name, conn)` pairs for agents to reap, close the map under the lock, then close connections after `h.mu.Unlock()` (or fire each `Close()` in its own goroutine). |
| Info | `services/agent-hub/internal/hub/hub.go:206-218` | `markOfflineIfCurrent` correctly closes the connection *after* releasing `h.mu.RLock()`, and only when the stored pointer still matches `expected`, so a stale liveness-check goroutine can't close a freshly re-registered agent's connection. No race identified here. | None — noted as good practice for contrast with the reaper finding above. |
| Info | `services/agent-hub/internal/hub/hub.go:181` | Pre-existing pattern (not introduced by this PR): `Register` closes the replaced connection while still holding `h.mu.Lock()`. Same class of issue as the reaper finding but out of scope for this diff. | Consider addressing together with the reaper fix in a follow-up. |
| Info | n/a | Close reason strings (`"liveness check failed"`, `"reaped: stale agent"`) sent to the peer are generic and disclose no sensitive internal state. | None. |

### Detail: reaper lock-holding Close (Medium)

Before this PR, the reaper never touched the socket, so holding `h.mu.Lock()` for the loop was cheap (pure map iteration). Now it performs blocking I/O (`conn.Close`) per stale entry inside the same critical section. Whether this is remotely triggerable as a cheap DoS depends on how easy it is for a caller to accumulate many stale/offline agents under the existing REST/WS auth middleware (out of scope for this diff to re-verify) — but independent of exploitability, this is a correctness/availability regression pattern worth fixing before it compounds: as more liveness-check paths funnel through the reaper, lock hold time grows with the number of concurrently-stale agents rather than being O(map ops).

I don't see this as attacker-controlled auth bypass or a new injection/deserialization vector — the change is narrowly scoped to closing an already-owned connection on transitions the hub itself decided (heartbeat timeout, keepalive failure, RPC-failure demotion, reap). Reconnection under the same name still goes through the existing `Register` path with its existing `AdvertiseAddress` check and pointer-identity guard in `markOfflineIfCurrent`; no new re-registration/authz gap identified.

**Recommendation:** Fix the reaper lock-holding `Close()` before merge (small, low-risk change — collect-then-close pattern). Not a hard blocker given it improves materially on the prior permanent-deadlock state, but should not ship as a follow-up ticket since the fix is cheap and touches the exact code this PR already modified.

No dependency, input-validation, or secrets-handling findings in this diff.

dokipen (commented):
## Test review: PR #790 (services/agent-hub/internal/hub/hub_test.go)

**Sanity checks performed:** `go test ./internal/hub/... -run 'TestHeartbeatTimeout|TestReaper' -v -race -count=3` and a follow-up `-run 'TestHeartbeatTimeoutClosesConnection|TestHeartbeatTimeoutThenReconnectRestoresOnline' -race -count=10`. All runs passed with the race detector enabled; no flakes observed.

### `TestHeartbeatTimeoutClosesConnection` (hub_test.go:247-320)
- Good: exercises the actual fix in `hub.go:210-215` — it uses `connectRawAgent` (a raw `*websocket.Conn`, not the echoing helper) so the heartbeat genuinely times out and `markOfflineIfCurrent` runs. It doesn't just check `agent.Status()`; it also proves the connection was closed by (a) attempting a post-offline write and confirming `LastSeen` doesn't advance, and (b) polling `conn.Read` for a close/error if the write raced the close handshake. This meaningfully falsifies the pre-fix behavior — before the fix, the write would have succeeded and `LastSeen` would have advanced, since the stale read loop stayed alive. Confirmed by inspecting the guarded `conn.Close(...)` call added at hub.go:213-215 and by re-running with `-race -count=10` (10/10 pass, no timing-based flake observed).
- Minor (non-blocking): the polling deadlines (2s for offline, 1s for the read-close check) are generous relative to the 50ms heartbeat interval/timeout used by `startTestHubWithHeartbeat`, so CI slowness is unlikely to cause false failures. No action needed.
- Minor (non-blocking): the double branch (write may or may not fail before poll) is a little verbose but well-commented and each branch has a clear assertion; leaving as-is is fine.

### `TestHeartbeatTimeoutThenReconnectRestoresOnline` (hub_test.go:407-441) + helpers `dialAndRegisterNoFatal` (hub_test.go:326-352) and `connectReconnectingAgent` (hub_test.go:359-401)
- Good: this is the acceptance-criteria test — it proves an agent can reconnect and come back online after a liveness-triggered offline transition, which is the actual user-facing behavior issue #691 cares about. Before the fix this test would hang/fail because the stale server-side read loop would keep running against the replaced map entry.
- `dialAndRegisterNoFatal` correctly avoids `t.Fatalf`/`t.FailNow` from a background goroutine (hub_test.go:326 doc comment states the rationale correctly — `t.FailNow` must only be called from the test's own goroutine). Good practice.
- Low-risk observation: `connectReconnectingAgent`'s background goroutine calls `t.Logf` on a failed reconnect (hub_test.go 391-ish, `t.Logf("reconnect failed: %v", err)`). `t.Logf` from a goroutine after the test function has returned will panic ("Log in goroutine after Test has completed"). Walking through the sequence, the reconnect+register must complete (synchronously, inside the goroutine) before the outer test's poll loop can observe `StatusOnline` via `h.Get`, so in the success path this log call cannot fire after the test returns. The risk window is only the reconnect-*failure* path, which would leave the test polling until its own 2s timeout and failing via `t.Fatalf` in the main goroutine first — so no double-completion race in practice. Still, this pattern (background goroutine touching `t` at all after registration) is worth a passing mention as a maintainability trap for whoever touches this test next; not a blocker.
- Coverage note: this test only exercises the `markOfflineIfCurrent` connection-close path (hub.go:213-215), not the reaper's connection-close path (hub.go:513-515).

### Reaper connection-close path — not covered
- Confirmed by reading `TestReaper` (hub_test.go:824-870): it registers agents with `conn == nil` (`h.Register("online-agent", nil, ...)`), so `agent.Conn()` is nil and the reaper's new `if conn := agent.Conn(); conn != nil { conn.Close(...) }` branch (hub.go:513-515) is never exercised — only the nil-skip path is. This matches the PR author's own note. The reaper's close call is a small, low-risk addition (same pattern as the already-tested `markOfflineIfCurrent` close, guarded by the same nil check), so the residual risk of it being wrong is low, but it is currently unverified by any test. Recommend a follow-up (not necessarily blocking this PR) that registers an agent with a real raw connection, marks it offline, waits for the reaper to reap it, and asserts the connection was closed (mirroring `TestHeartbeatTimeoutClosesConnection`'s write/read-based proof). This would close the one meaningful coverage gap in this PR.

### Overall
- Tests are well-targeted at the actual fix (not incidentally passing) and are low-flakiness (verified via `-race -count=10`).
- No blocking issues found in the tests as submitted.
- Suggested but non-blocking follow-up: add a reaper-specific connection-close test to cover hub.go:513-515 (currently only reachable via nil-conn agents in `TestReaper`).

dokipen (commented):
## Code Review

| Warning | `services/agent-hub/internal/hub/hub.go:509-520` | Reaper calls `conn.Close(...)` for each stale agent **while holding `h.mu.Lock()`** (the hub's global map mutex, used by `Register`, `Get`, list/RLock paths throughout the file). `coder/websocket`'s `Close` performs a close handshake with up to a 5s timeout waiting on the peer (`close.go:185`), so a single unresponsive stale connection blocks *every* hub operation needing `h.mu` for up to 5s — and this is serial per agent, so N stale agents in one reaper tick can block the hub for N×5s. | Mirror `markOfflineIfCurrent`'s pattern (hub.go:206-218): collect the `(name, conn)` pairs to reap and delete from the map while holding the lock, then `Unlock()` and close the connections outside the lock. |
| Suggestion | `services/agent-hub/internal/hub/hub_test.go:824` (`TestReaper`) | The reaper fix (closing the stale agent's conn before eviction) has no test coverage — `TestReaper` registers agents with a `nil` conn, so the new `agent.Conn() != nil` / `conn.Close(...)` branch in the reaper is never exercised by any test. Only the heartbeat-timeout path (`markOfflineIfCurrent`) is covered by the two new tests. | Add a reaper test that registers an agent with a real (test-server) websocket connection, marks it offline, and asserts the connection is closed by the time it's reaped — analogous to `TestHeartbeatTimeoutClosesConnection`. |

### Details

**Reaper lock scope (Warning, does not strictly block merge but should be tracked):** Holding `h.mu.Lock()` across a network call with an internal multi-second timeout is a latent availability risk — under normal conditions (peer already gone, TCP RST) `Close` returns quickly, but a half-open/stalled peer connection can stall the entire hub's registration and lookup paths. This is inconsistent with `markOfflineIfCurrent`, which already does the right thing by releasing `h.mu` (via `RUnlock`) before calling `Conn().Close()`. Given `#691`'s whole point is fixing a liveness deadlock, introducing a new (much shorter-lived, but real) lock-contention window in the reaper is worth tightening before this ships, or at minimum tracked immediately after.

### Verified

- `conn.Close()` is safe to call twice (`close.go:102` CAS guard, second call is a no-op returning `net.ErrClosed` which is squashed to nil) — the double `"agent marked offline"` log line seen in test output (once from the RPC/heartbeat path, once from the read loop exiting after `Read` errors) is expected and harmless, just log noise.
- The nil-guard on `expected.Conn()` is **not** dead code — `TestReaper` (hub.go pre-existing test) registers agents with a `nil` conn via `h.Register(name, nil, ...)`, so the guard is exercised and necessary, not defensive-only.
- `go test ./internal/hub/... -race -count=1` passes; the two new tests (`TestHeartbeatTimeoutClosesConnection`, `TestHeartbeatTimeoutThenReconnectRestoresOnline`) pass individually and are well-targeted at the deadlock scenario from #691 (verify both that the old connection stops being usable/touched and that reconnection under the same name restores online status).

**Deferred**: Tighten reaper lock scope to close connections outside `h.mu` → same ticket family as #691, recommend a follow-up low/medium priority ticket if not fixed in this PR, discovered in #790 review.

REVIEW STATUS: CHANGES NEEDED
- Move `conn.Close()` calls in the reaper (hub.go:514-516) outside the `h.mu.Lock()` critical section to avoid blocking the hub on a stalled peer's close handshake.

dokipen:
Fixed the review feedback: the reaper now collects stale agents' connections into a slice while holding `h.mu.Lock()`, releases the lock, then closes each connection outside the critical section — mirroring `markOfflineIfCurrent`'s existing pattern. See `3c5704f`.

Verified: `go build`, `go vet`, `gofmt -l` clean; `TestReaper -race -count=5` and full `internal/hub` suite with `-race` pass.

